PR #14, opened 2016-07-19
Useful error descriptions

Stebalien:
"unknown error" is mildly rage inducing.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -35,7 +35,11 @@ impl error::Error for Error {
             Error::Syntax => "syntax error",
             Error::Io(ref error) => error::Error::description(error),
             Error::FromUtf8(ref error) => error.description(),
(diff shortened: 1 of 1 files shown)


Stebalien:
Note: The CI fail is unrelated.

pyfisch:
Thank you for improving error handling. The PR is now added to master. Do you think it is worth it to add variants for [all other error types](https://serde-rs.github.io/serde/serde/de/trait.Error.html) to the cbor type, or is it ok to map them all to custom error?

Stebalien:
It would be nice but I'd say that's pretty low priority. IMO, one shouldn't even try to automatically recover from (de)serialization errors. The only errors one needs to be able to distinguish between are IO/non-IO errors.

I only made this pull request to make debugging easier.

PR #15, opened 2016-07-22
Don't serialize newtyped types as arrays

Stebalien:
This is a *breaking change* (mirroring https://github.com/serde-rs/json/pull/115).


pyfisch:
Merged with test.

Stebalien:
Thanks!

PR #25, opened 2017-02-03
Fix handling of variable length maps and lists

### LINKED ISSUE #24: Packed deserialization changes broke some kinds of deserialization
This input is correct CBOR that is deserialized in 0.5.0, but not 0.5.1:

```
bf 67 6d 65 73 73 61 67 65 64 70 6f 6e 67 ff
```

It should translate to

```json
{
    "message": "pong"
}
```

Fails at I believe this line with a value of 7: https://github.com/pyfisch/cbor/blob/master/src/de.rs#L296

sfackler:
This feels a bit gross - is there a nicer way of handling stop words?

Closes #24


pyfisch:
Thank you!

Issue #17, opened 2016-07-28
Support the canonical format.


Stebalien:
See https://tools.ietf.org/html/rfc7049#section-3.9.

This makes it possible to compare objects for bitwise equality (it also makes signatures/signature verification easier).

pyfisch:
These are the requirements for canonical format. 

* [x] Integers must be as small as possible.
* [x] The expression of lengths in major types 2 through 5 must be as
      short as possible.  The rules for these lengths follow the above
      rule for integers.
* [ ] The keys in every map must be sorted lowest value to highest.
      Sorting is performed on the bytes of the representation of the key
      data items without paying attention to the 3/5 bit splitting for
      major types.
* [ ] Indefinite-length items must be made into definite-length items.

chrysn:
The way item sorting is currently implemented makes solving #79 a bit harder as it places serialized objects in temporarily allocated memory. That's in theory a non-issue when serializing BTreeMap data (which won't be there in embedded anyway -- it's only about avoiding ifdef hell there), but harder AFAICT for serialization of non-packed struct fields (as serde calls serialize_field, so we can't apply the field ordering at build time).

Is there currently (or would it make sense to have) an option to disable canonicalization in a serializer? (I figure it may have performance benefits on std as well, but can't tell whether that's actually an issue).

chrysn:
Two more notes in this context:
* The online reference implementation does not produce canonical CBOR either, weakly indicating that we might not need to produce canonical CBOR unconditionally (but only when requested).
* There are some [pending errata items on the CBOR RFC](https://www.rfc-editor.org/errata_search.php?rfc=7049) that relate to canonicalization. There is an update to CBOR [in the works](https://tools.ietf.org/html/draft-ietf-cbor-7049bis-04) that addresses those. (Which is not really breaking, given that the canonicalization rules are suggestions for what particular applications could specify as their canonical format, rather than prescribing what canonical CBOR is).

pyfisch:
This is done. Convert the values you want to serialize in canonical format to values and then call serialize.

PR #35, opened 2017-06-09
Change Value::Object to BTreeMap

jethrogb:
Rust's `HashMap`'s iteration order is non-deterministic, meaning deserializing into `Value` and then serializing has a non-deterministic output. This PR changes that by storing a `Value::Object` as a `BTreeMap`, the same way `serde_json` does it. This is a [breaking-change].

Note that this PR does not preserve the order of incoming CBOR, it merely makes sure that the output order is always the same. `serde_json` has an optional `preserve_order` feature using [linked-hash-map](https://crates.io/crates/linked-hash-map) which would need to be implemented to reach parity with `serde_json`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -94,11 +94,11 @@
 //! Read a CBOR value that is known to be a map of string keys to string values and print it.
 //!
 //! ```rust
-//! use std::collections::HashMap;
+//! use std::collections::BTreeMap;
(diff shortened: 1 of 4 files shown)


jethrogb:
Possible step towards #17

pyfisch:
Thank you.

PR #36, opened 2017-06-16
Fix tuple deserialization from indefinite-length array

padamowicz:
Serde only calls `next_element` n times for a tuple of n elements (https://github.com/serde-rs/serde/blob/master/serde/src/de/impls.rs#L749), therefore we never consume the delimiter when deserializing into tuples from indefinite-length arrays.

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -222,8 +222,6 @@ impl<'de, 'a, R: Read> de::Deserializer<'de> for &'a mut Deserializer<R> {
         deserialize_byte_buf();
         deserialize_map();
         deserialize_unit_struct(_name: &'static str,);
-        deserialize_tuple_struct(_name: &'static str, _len: usize,);
-        deserialize_tuple(_len: usize,);
         deserialize_ignored_any();
     );
     #[inline]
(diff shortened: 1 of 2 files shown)


src/de.rs
```diff
@@ -344,6 +342,38 @@ impl<'de, 'a, R: Read> de::Deserializer<'de> for &'a mut Deserializer<R> {
             .expect("struct_fields is not empty");
         Ok(result)
     }
+
+    #[inline]
+    fn deserialize_tuple<V: Visitor<'de>>(self, _len: usize, visitor: V) -> Result<V::Value> {
```
sfackler: I think this logic needs to be handled in the deserialize_any codepath - we want this check to happen even if someone calls `deserialize_any` rather than `deserialize_tuple`.

pyfisch:
Hi @padamowicz , can you please explain your code a bit more? And respond to @sfackler's comment?

sfackler:
I have a larger rewrite that also picks up things like zero copy deserialization that I need to finish off (the compact encoding format needs to be finished I think) that also fixes these kinds of issues in a more "first class" way. https://github.com/sfackler/cbor/tree/rewrite

pyfisch:
@sfackler the rewrite is great. Resolving these issues in a "first class" way is important because after fixing they frequently reappear at a different place.

PR #40, opened 2017-07-12
Allows to customize sequence limit when deserializing (fixes #39)

### LINKED ISSUE #39: What's the purpose of MAX_SEQ_LEN ?
I need to serialize bytes longer than 0.5Mb. Can the constant be customized or just dropped?

tailhook:
Also, it's a good idea to make some better error message when the limit is reached because "syntax error" is misleading.

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -486,6 +486,8 @@ impl<'de, 'a, R: Read> de::VariantAccess<'de> for CompositeVisitor<'a, R> {
 }
 
 /// Decodes a CBOR value from a `std::io::Read`.
+///
+/// Note: this function decodes with default value of `set_sequence_limit`
 #[inline]
 pub fn from_reader<T: DeserializeOwned, R: Read>(reader: R) -> Result<T> {
     let mut de = Deserializer::new(reader);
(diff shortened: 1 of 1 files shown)


tailhook:
The second commit adds a sensible limit for `from_slice` in my opinion. Feel free to apply only the first one if you feel the other way.

src/de.rs
```diff
@@ -81,7 +102,9 @@ impl<'de, R: Read> Deserializer<R> {
     fn parse_size_information(&mut self, first: u8) -> Result<Option<usize>> {
         let n = self.parse_additional_information(first)?;
         match n {
-            Some(n) if n > MAX_SEQ_LEN => return Err(Error::Syntax),
+            Some(n) if n > self.max_seq_len as u64 => {
```
pyfisch: This should only apply to strings and bytestrings not all sequences.

tailhook: Hm, I wanted to ask why. But I see that in serde that `size_hint` is already capped at 4096, so yes it might not make sense for sequences. And with this in mind, your proposal of growable buffer for strings and bytes is simpler to implement too.

tailhook:
Closing in favor of #41

PR #41, opened 2017-07-18
Got rid of MAX_SEQ_LEN (fixes #39)

### LINKED ISSUE #39: What's the purpose of MAX_SEQ_LEN ?

tailhook:
Another attempt to fix #39. Looks easy. I'm not sure about initial vector capacity, but otherwise patch is quite simple.

Note: serde itself caps `size_hint` to 4096 when serializing (i.e. doesn't allocate more than 4096 items up front)

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -1,14 +1,14 @@
 //! CBOR deserialization.
 
 use std::io::{self, Read};
+use std::cmp::min;
+use std::usize;
 
 use byteorder::{BigEndian, ReadBytesExt};
 use serde::de::{self, Visitor, Deserialize, DeserializeOwned, DeserializeSeed};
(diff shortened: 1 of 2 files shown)


src/de.rs
```diff
@@ -81,7 +81,7 @@ impl<'de, R: Read> Deserializer<R> {
     fn parse_size_information(&mut self, first: u8) -> Result<Option<usize>> {
         let n = self.parse_additional_information(first)?;
         match n {
```
pyfisch: Remove this match. n is always smaller i64 max.

pyfisch: Ah it is for 32bit architecture.

Issue #3, opened 2016-02-06
Parsing and Emitting Tags


vooon:
Hello. I'm what to emit Tag(0) (RFC 3339 date-time string), but i can't find any hints.

pyfisch:
It is currently not possible because serde does not support serializing and deserializing tags. If you need it I will propose a PR to serde to add this function.

vooon:
Thank you. But for my "hello world" class project easier to use *go* or even *c11*.

dignifiedquire:
@pyfisch I'm looking into using this for an implementation, but I will need to be able to use custom tags. How far of is that? Or should I just use http://burntsushi.net/rustdoc/cbor/ for now, which already implements handling of custom tags, though as I'm using serde_json already in other parts I'd really like to use serde_cbor.

pyfisch:
Adding support for tags to serde_cbor requires support from the serde crate. There is already an issue about adding adding tags to serde: serde-rs/serde/issues/163 but it has not yet been decided how the support should look like. Tags are specific to serializiation formats and for this reason it is hard to handle them in a generic serializiation framework.

@erickt Do you have any ideas about this?

dvc94ch:
Any news on this?

pyfisch:
Nope, if this should be done well it needs some help from serde. Currently there is a lot of work going on related to no-std compatibility and I plan to generally polish the crate. When this is finished tags can be discussed again.

sfackler:
This may need to wait for specialization, where this crate could define subtraits of Serializer/Deserializer/etc that provide extra methods to interact with tags. Serialize/Deserialize implementations could then specialize for that subtrait to interact with the extra functionality.

petreeftime:
I'm also interested in tagging. I would like to implement a COSE library [https://tools.ietf.org/html/rfc8152], but that requires tags. https://github.com/BurntSushi/rust-cbor suggests using this crate, since that one is unmaintained, but this one doesn't have tags. 

SerDe seems stuck on tagging due to missing specialization support: https://github.com/serde-rs/serde/pull/455. Do you think adding some extra tagging API to this crate in the interim could make this work?

PR #43, opened 2017-09-16
Major rewrite

### LINKED ISSUE #31: Remove dependency on serde_bytes
Parse bytestrings of indefinite length [without using the ByteBuf wrapper](https://github.com/pyfisch/cbor/blob/e3bacca2a1765efe682926aed775e9bf65ce4715/src/de.rs#L123).

Also refactor parsing indefinite length UTF8-strings.

### LINKED ISSUE #32: Support zero copy deserialization

### LINKED ISSUE #37: Packed encoding for structs is broken when fields are conditionally skipped
Because serde doesn't provide field indices when serializing, the serializer currently guesses the index based off of the number of fields serialized. This doesn't work when fields are conditionally skipped, however:

```rust
extern crate serde;
extern crate serde_cbor;
#[macro_use]
extern crate serde_derive;

#[derive(Serialize, Deserialize)]
struct Foo {
    a: u32,
    #[serde(skip_serializing_if = "Option::is_none")]
    b: Option<u32>,
    c: u32,
}

#[test]
fn foo() {
    let foo = Foo {
        a: 0,
        b: None,
        c: 1,
    };
    let buf = serde_cbor::ser::to_vec_packed(&foo).unwrap();
    serde_cbor::from_slice::<Foo>(&buf).unwrap();
}
```

```
running 1 test
test foo ... FAILED

failures:

---- foo stdout ----
	thread 'foo' panicked at 'called `Result::unwrap()` on an `Err` value: Custom("missing field `c`")', /checkout/src/libcore/result.rs:859
note: Run with `RUST_BACKTRACE=1` for a backtrace.


failures:
    foo

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured

error: test failed, to rerun pass '--lib'

Completed with code 101
```

sfackler:
This commit consists of a rewrite of basically all of the serialization
and deserialization logic. The API and internals are based heavily off
of serde_json, in particular, the error type and support for borrowed
deserialization. The deserialization logic is now based off of the tag
byte lookup table described in appendix B of RFC 7049. It's a bit more
verbose, but compiles down to a jump table so there should be some
decent performance benefits in deserialization. It should also resolve
some longstanding correctness issues with deserialization of indefinite
length maps and sequences. The packed encoding can now properly handle
conditionally skipped fields and doesn't need special logic on the
deserialization side due to new serde features.

Closes #37
Closes #36
Closes #32
Closes #31

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -549,7 +549,7 @@ where
                 self.parse_u64()?;
                 self.parse_value(visitor)
             }
-            0xde...0xdf => Err(self.error(ErrorCode::UnassignedCode)),
+            0xdc...0xdf => Err(self.error(ErrorCode::UnassignedCode)),
 
             // Major type 7: floating-point numbers and other simple data types that need no content
             0xe0...0xf3 => Err(self.error(ErrorCode::UnassignedCode)),
--- a/src/de.rs
+++ b/src/de.rs
@@ -1,479 +1,960 @@
-//! CBOR deserialization.
+//! Deserialization.
 
-use std::io::{self, Read};
-use std::cmp::min;
-use std::usize;
+use byteorder::{ByteOrder, BigEndian};
+use serde::de;
+use std::io;
+use std::str;
+use std::f32;
+use std::result;
 
-use byteorder::{BigEndian, ReadBytesExt};
-use serde::de::{self, Visitor, Deserialize, DeserializeOwned, DeserializeSeed};
-use serde_bytes::ByteBuf;
+use error::{Error, Result, ErrorCode};
+use read::Reference;
+pub use read::{Read, IoRead, SliceRead};
 
-use super::error::{Error, Result};
+/// Decodes a value from CBOR data in a slice.
+pub fn from_slice<'a, T>(slice: &'a [u8]) -> Result<T>
+where
+    T: de::Deserialize<'a>,
+{
+    let mut deserializer = Deserializer::from_slice(slice);
+    let value = de::Deserialize::deserialize(&mut deserializer)?;
+    deserializer.end()?;
+    Ok(value)
+}
 
-macro_rules! forward_deserialize {
-    ($($name:ident($($arg:ident: $ty:ty,)*);)*) => {
-        $(#[inline]
-        fn $name<V: Visitor<'de>>(self, $($arg: $ty,)* visitor: V) -> Result<V::Value> {
-            self.deserialize_any(visitor)
-        })*
+/// Decodes a value from CBOR data in a reader.
+pub fn from_reader<'a, T, R>(reader: R) -> Result<T>
+where
+    T: de::Deserialize<'a>,
+    R: io::Read,
+{
+    let mut deserializer = Deserializer::from_reader(reader);
+    let value = de::Deserialize::deserialize(&mut deserializer)?;
+    deserializer.end()?;
+    Ok(value)
+}
+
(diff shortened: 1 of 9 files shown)


src/de.rs
```diff
@@ -1,479 +1,960 @@
...
+                let len = self.parse_u64()?;
+                if len > usize::max_value() as u64 {
+                    return Err(self.error(ErrorCode::LengthOutOfRange));
+                }
+                self.parse_bytes(len as usize, visitor)
+            }
+            0x5c...0x5e => Err(self.error(ErrorCode::UnassignedCode)),
+            0x5f => self.parse_indefinite_bytes(visitor),
+
+            /// Major type 3: a text string
```
pyfisch: Use normal comment.

src/de.rs
```diff
@@ -1,479 +1,960 @@
...
-            _ => unreachable!(),
+    fn next(&mut self) -> Result<Option<u8>> {
+        self.read.next().map_err(Error::io)
+    }
+
+    fn peek(&mut self) -> Result<Option<u8>> {
+        self.read.peek().map_err(Error::io)
+    }
+
+    fn eat(&mut self) {
```
pyfisch: Maybe `fn consume` like in `BufReader`?

src/error.rs
```diff
@@ -1,92 +1,141 @@
...
-            Error::SizeError => "size error",
-            Error::Io(ref error) => StdError::description(error),
-            Error::FromUtf8(ref error) => error.description(),
-            Error::Custom(ref s) => s,
-            Error::StopCode => "unexpected stop code",
-            Error::TrailingBytes => "unexpected trailing bytes",
-            Error::__Nonexhaustive => unreachable!(),
+        match self.0.code {
+            ErrorCode::Io(ref err) => error::Error::description(err),
+            _ => "JSON error",
```
pyfisch: "CBOR error"

jq-rs:
Looks fantastic! Maybe 1.0 after these changes? If so, maybe we can get rid of unmaintained on the crate description after 1.0 as it really scares everybody?

pyfisch:
> Maybe 1.0 after these changes?
Issue #3 is still open. I would like to see this fixed but it does not seem to be a priority for serde itself (and might even require a breaking release for the main serde crate). Before 1.0 the crate should also be fuzzed again and used for some time.

> If so, maybe we can get rid of unmaintained on the crate description after 1.0 as it really scares everybody?

If I can find a trustworthy person wanting to maintain the crate the unmaintained can be removed. dtolnay rejected the idea to move the crate to the @serde-rs organization so it can't probably become an official serde crate. Maybe @sfackler wants to adopt the crate but he did not say so yet. (If no one is found to adopt the crate we can discuss to make the unmaintained label smaller as most long-standing issues are resolved with this rewrite and I will still review issues and PRs)

sfackler:
Yeah, I'd be happy to adopt the crate if that'd help with things!

The other big change that I'd want to make before a 1.0 is to overhaul Value to more fully support everything in CBOR. There are some tricks that the toml crate pulled with TOML datetime values that we can hopefully use here to get tags and things like that out in a useable way.

sfackler:
Updated!

pyfisch:
Thanks a lot!

arthurprs:
This is great work.

I think the readme unmaintained alert will throw a lot of people off though, maye remode it?

PR #44, opened 2017-09-24
Add fuzz targets for from_slice and from_reader

sfackler:
These found the issues fixed in https://github.com/pyfisch/cbor/pull/43/commits/6d23f781cbef0f99b27b4899e3053561d837f99e and https://github.com/pyfisch/cbor/pull/43/commits/8f0079129c6a4f7232761ffc3de452c9a67c81ae.

I've also included a script to initialize a corpus based off of the reference examples in Appendix A of the RFC. The fuzzer appears to do ~50% better with coverage staring off of that base rather than nothing.

### DIFF
--- a/fuzz/.gitignore
+++ b/fuzz/.gitignore
@@ -1,0 +1,4 @@
+
+target
+corpus
+artifacts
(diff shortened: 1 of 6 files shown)


pyfisch:
Thanks. I will release the new version soon.

PR #49, opened 2017-10-19
Implement to_value serialization

mersinvald:
I have ported `to_value` serialization from `serde_json` in the way that would keep `Value` structure high-level, using `Object` where it was possible instead of `Bytes` sequences as the default `cbor` serializer does when it outputs the serialized data into a byte stream.

This makes impossible to perform `Object -> Value -> SerializedString -> Object` conversion without intermediate deserialization to `Value` because layouts of initial `Object` and the `Value` are incompatible, but it keeps the `Value` useful for inspection and conversions.

The `Value` with byte streams instead of the most of complex types is not very useful.


pyfisch:
@sfackler What do you think?

anton-dutov:
@pyfisch
ping

PR #54, opened 2017-12-25
Set Serializer is_human_readable to false.

### LINKED ISSUE #51: Override `is_human_readable`
According to the docs when [`is_human_readable`](https://docs.rs/serde/1.0.23/serde/trait.Serializer.html#method.is_human_readable) is `false` serializer can produce more compact and efficient form. I think this is what expected from cbor serializer. (default is `true`)

Unfortunately changing it is a breaking change, but I think it's okay in zero-point release.

So I'll make a PR, if you're fine with it?

Alternative would be to make it configurable, but I don't think human-readable things in binary cbor is a good default.

pyfisch:
Closes #51

BREAKING CHANGE: This results in a different serialization
for common types e.g. Ipv4Addr.

### DIFF
--- a/src/ser.rs
+++ b/src/ser.rs
@@ -453,6 +453,10 @@ where
         self.serialize_unit_variant(name, variant_index, variant)?;
         self.serialize_struct(name, len)
     }
+
+    fn is_human_readable(&self) -> bool {
+        false
+    }
 }
 
 impl<'a, W> ser::SerializeTuple for &'a mut Serializer<W>
--- a/tests/ser.rs
+++ b/tests/ser.rs
@@ -91,3 +91,13 @@ fn test_self_describing() {
     }
     assert_eq!(vec, b"\xd9\xd9\xf7\x09");
 }
+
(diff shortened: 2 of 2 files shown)


sfackler:
Do you think it's worth making this configurable?

pyfisch:
I am not sure. Maybe there are people who need/want to retain the old behavior, but I don't know why.

sfackler:
I could see it potentially causing issues when trying to interoperate with other CBOR processors. We can always add it later though if it ends up being a problem.

RReverser:
This actually broke deserialization of Ipv4Addr - Deserializer must have `is_human_readable` returning the same value. Right now, if you try to serialize Ipv4Addr (or, really, any other human-readability-dependant format), it will serialize it as sequence of bytes, but will try to deserialize as a string, resulting in an error:

```
called `Result::unwrap()` on an `Err` value: ErrorImpl { code: Message("invalid type:sequence, expected a string"), offset: 0 }
```

PR #56, opened 2018-01-14
Add support for CBOR semantic tags

TethysSvensson:
Hi,

I've tried to add support for CBOR tags. It is **pretty** hacky and feel free to reject it if you don't think this is the right approach. I'm not actually convinced that this is a good idea at all, but I wanted to try hacking it in. And now that I am done, I might as well ask for your opinion.

If rust-lang/rust#41875 is ever implemented, this code should become a lot more safe and easier to read. In the meantime, here we are.

### DIFF
--- a/src/ser.rs
+++ b/src/ser.rs
@@ -1,7 +1,7 @@
 //! Serialize a Rust data structure to CBOR data.
 use byteorder::{ByteOrder, BigEndian};
 use serde::ser::{self, Serialize};
-use std::io;
+use std::{io, mem};
 
 use error::{Error, Result};
 
@@ -411,10 +411,22 @@ where
     #[inline]
     fn serialize_tuple_struct(
         self,
-        _name: &'static str,
+        name: &'static str,
         len: usize,
     ) -> Result<&'a mut Serializer<W>> {
-        self.serialize_tuple(len)
+        use tagged::CBOR_NO_LENGTH_FIELD;
+        if name.as_ptr() == CBOR_NO_LENGTH_FIELD.as_ptr() && name == CBOR_NO_LENGTH_FIELD {
+            // We put both the pointer comparison and the string comparison, in case the optimizer
+            // decides to do something funny and make two different strings have the same pointer.
+            //
+            // It might be possible to intentionally get into this branch be a user of the library,
+            // but only if they name their tuple-struct "__CBOR_NO_LENGTH_FIELD", which will not
+            // happen accidentally. If they do, then the only thing that will accomplish is to make
+            // us generate invalid CBOR.
+            Ok(self)
+        } else {
+            self.serialize_tuple(len)
+        }
     }
 
     #[inline]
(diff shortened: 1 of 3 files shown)


src/ser.rs
```diff
@@ -411,10 +411,22 @@ where
     #[inline]
     fn serialize_tuple_struct(
         self,
-        _name: &'static str,
+        name: &'static str,
         len: usize,
     ) -> Result<&'a mut Serializer<W>> {
-        self.serialize_tuple(len)
+        use tagged::CBOR_NO_LENGTH_FIELD;
+        if name.as_ptr() == CBOR_NO_LENGTH_FIELD.as_ptr() && name == CBOR_NO_LENGTH_FIELD {
```
sfackler: What's the point of comparing the pointers?

src/ser.rs
```diff
@@ -492,7 +504,21 @@ where
     where
         T: ?Sized + ser::Serialize,
     {
-        value.serialize(&mut **self)
+        use tagged::CborTag;
+        let t_fn: fn(&T, Self) -> Result<()> = T::serialize;
```
sfackler: This seems like an unreliable way of doing this detection. You could e.g. potentially end up with a situation where there are two copies of `CborTag::serialized` that ended up in different compilation units.

sfackler:
@dtolnay is there a standard way to do this kind of thing?

dtolnay:
I don't think there is a standard way but you need to do it above-board, meaning that the Rust representation of a tagged value serializes as some pattern in the data model and the Serializer recognizes that pattern and emits it as a CBOR tagged value. This approach would avoid the pointer comparison funny business. For example you could decide that a tagged value in the data model is represented as a tuple struct with the name "serde_cbor::Tag" where the first element is the u64 tag and the second element is any value T.

src/ser.rs
Idolf: My primary goal was to avoid external users from getting inside the `if` statement. Secondary goal was speed. After having slept on it, the first seems like a stupid (and unobtainable) goal, while the second will probably not matter at all.

Idolf:
I agree that the pointer comparisons are pretty ugly. I mostly coded this, just to see if it could be done.

So the issue is that, an instance of Serialize would like to be sure that **only** the CBOR Serializer will change behavior. This means that we cannot use a tuple struct without knowing if you are serializing for CBOR or not.

I can see two less hacky approaches to this:
* Add "type Marker: Any" to the Serializer trait. That way you can detect if you are trying to emit to the CBOR serializer. Unfortunately this is a breaking change in serde (unless we wait for `associated_type_defaults`).
* Add this to the Serializer trait:
```rust
fn serialize_with_hint<H, T>(self, hint: &H, value: &T) -> Result<Self::Ok, Self::Error>
    where
        H: Any,
        T: Serialize;
```

This can be done without being a breaking change, since an obvious default implementation is possible.

pyfisch:
I did not react to this PR yet but I am very interested in supporting tagged values in this crate.

pyfisch:
Gone stale, I will keep this PR in mind when tags are added.

PR #60, opened 2018-02-22
WIP: add support for parsing stringref

tailhook:
Hi,

My intention is to support [stringref](http://cbor.schmorp.de/stringref) spec for cbor. The basic idea of the spec is to reference duplicate strings instead of copying them many times. It has two tags:

1. Start the namespace (start counting strings from zero) tag 256
2. Reference a string tag 25

As the spec says, this functionality must specifically be supported by encoder/decoder because you need to enumerate every non-tagged string in the right order to be able to refer to it later.

Currently, this PR implements a very minimal (and not fully compliant) decoder just to show the impact on the codebase of the crate. The functionality is entirely disabled (only empty vec is initialized) until the tag 256 appears so I believe it doesn't affect performance.

While currently this spec is not widely supported the tag numbers are reserved by IANA so will not conflict with anything. And I hope to promote and increase adoption of this spec by other implementations too.

What do you think? Will this functionality be accepted into serde_cbor?


tailhook:
Well, I'm going to defer this work. Several use-cases I've tested are better compressed by gzip. So I'm closing this PR until a good use case could be found.

PR #62, opened 2018-04-09
Document using serde_bytes with CBOR byte strings

### LINKED ISSUE #53: `[u8; 4]` seralized as array of 4 numbers instead of byte string.
How do I use byte strings with serde_cbor?

I want `"64.71.168.211".parse::<std::net::Ipv4Addr>().unwrap().octets()` to be `54 40 47 a8 d3`, not   `84 18 40 18 47 18 a8 18 d3`.

asomers:
CBOR byte strings are much more efficient than generic arrays.  Also,
add tests for encoding CBOR's various fixed-width byte string types.

Fixes #53

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -125,6 +125,20 @@
 //! let encoded = to_vec(&programming_languages);
 //! assert_eq!(encoded.unwrap().len(), 103);
 //! ```
+//!
+//! Serializing a `Vec` as a specialized byte string uses about 2x less RAM and
+//! 100x less CPU than as a default array.
+//! ```rust
+//! use std::collections::BTreeMap
+//! use serde_bytes::ByteBuf;
+//! use serde_cbor::to_vec;
+//!
+//! let data: Vec<u8> = vec![0, 1, 255];
+//! let byte_buf = ByteBuf::from(data);
+//! let serialized_array = to_vec(&data);
+//! let serialized_byte_string = to_vec(byte_buf);
+//! assert!(serialized_byte_string.len() < serialized_array.len());
+//! ```
 
 #![deny(missing_docs)]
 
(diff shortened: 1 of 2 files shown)


asomers:
Something is sending `SIGKILL` to the test process.  Does Travis object to tests that use 8GB of RAM?  I'll disable that part of the test.

sfackler:
Travis nodes have 7.5 GB of RAM.

tests/ser.rs
```diff
@@ -103,3 +105,40 @@ fn test_ip_addr() {
...
+    assert_eq!(&long_s[0..3], [0x59, 1, 0]);
+    assert_eq!(&long_s[3..], &long[..]);
+
+    let very_long_vec = (0..65536).map(|i| (i & 0xFF) as u8).collect::<Vec<_>>();
+    let very_long = ByteBuf::from(very_long_vec);
+    let very_long_s = to_vec(&very_long).unwrap();
+    assert_eq!(&very_long_s[0..5], [0x5a, 0, 1, 0, 0]);
+    assert_eq!(&very_long_s[5..], &very_long[..]);
+
+    // This assertion takes about 8GB of RAM, so Travis kills it :(
```
pyfisch: Can you please remove the commented out test? If it is not run by default it is very likely to become broken by future updates without anyone noticing.

src/lib.rs
```diff
@@ -125,6 +125,25 @@
 //! let encoded = to_vec(&programming_languages);
 //! assert_eq!(encoded.unwrap().len(), 103);
 //! ```
+//!
+//! Serializing a `Vec` as a specialized byte string uses about 2x less RAM and
+//! 100x less CPU than as a default array.
```
pyfisch: Can you be more specific here? I think you mean "less CPU time" but I was a little confused when reading.

PR #64, opened 2018-04-11
de::from_reader can't return borrowed data

### LINKED ISSUE #63: `de::from_reader` should implement `DeserializeOwned`, not `Deserialize`
I tried to use `serde_cbor::de::from_reader` to deserialize into a `&str` and got the surprising runtime error message like this:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ErrorImpl { code: Message("invalid type: string \"foobar\", expected a borrowed string"), offset: 0 }', libcore/result.rs:945:5
```
It makes sense that `from_reader` can't do zero-copy like `from_slice` since there's no source buffer.  However, attempting to do so should cause a compile-time failure, not a runtime failure.  docopt had a similar bug: https://github.com/docopt/docopt.rs/pull/222

asomers:
de::from_reader can't do zero-copy like de::from_slice can.  Fix its
type bounds so that attempts to use it for zero-copy will fail at
compile-time.

Fixes #63

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -12,6 +12,26 @@ use read::Reference;
 pub use read::{Read, IoRead, SliceRead};
 
 /// Decodes a value from CBOR data in a slice.
+///
+/// # Examples
+///
+/// Deserialize a `String`
+///
+/// ```
+/// # use serde_cbor::de;
+/// let v: Vec<u8> = vec![0x66, 0x66, 0x6f, 0x6f, 0x62, 0x61, 0x72];
+/// let value: String = de::from_slice(&v[..]).unwrap();
+/// assert_eq!(value, "foobar");
+/// ```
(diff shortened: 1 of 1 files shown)


sfackler:
This seems reasonable to me but it is a breaking change. I'd probably hold off on it until we have more changes we'd like to fold in.

dtolnay:
This working code would be broken.

```rust
fn my_cbor_helper<'de, T, R>(reader: R) -> serde_cbor::Result<T>
where
    T: Deserialize<'de>,
    R: Read,
{
    serde_cbor::from_reader(reader)
}

let s: String = my_cbor_helper(...).unwrap();
```

pyfisch:
I will merge this with the next 0.x release.

PR #66, opened 2018-05-10
implement from_value() to convert Value to any type implementing Deserialize

astraw:
Hi, this implements `fn from_value<T: DeserializeOwned>(value: Value) -> Result<T, Error>` which takes a `serde_cbor::Value` and converts it to type `T`. This is much like the equivalent function in serde_json.

The implementation is very simple but also suboptimal - it serializes `serde_cbor::Value` and then deserializes into type `T`. I thought I would propose this implementation as a start, because I did not dig in enough to check if there was already substantial machinery in place to perform this conversion directly or if an implementation of `serde::de::Deserializer` for `serde_cbor::Value` would need to be written from scratch so that `T::deserialize(value)` would directly work.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -144,4 +144,4 @@ pub use de::{from_slice, from_reader, Deserializer};
 #[doc(inline)]
 pub use ser::{to_writer, to_vec, Serializer};
 #[doc(inline)]
-pub use value::{Value, ObjectKey, to_value};
+pub use value::{Value, ObjectKey, to_value, from_value};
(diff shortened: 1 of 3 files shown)


sfackler:
Seems reasonable to me. I'd look into what serde_json does to implement its version of this method. If there was existing machinery, it'd probably use it.

astraw:
serde_json uses `T::deserialize(value)`, but this depends on `serde::de::Deserializer` being implemented for `serde_json::Value`. Since `serde::de::Deserializer` is a somewhat big trait, the implementation required is non-trivial (unless it already more-or-less exists in the serde_cbor code base, which is what I did not dig in enough to find).

astraw:
Great, thanks.

PR #72, opened 2018-07-14
Add a StreamDeserializer

### LINKED ISSUE #68: Detecting invalid cbor when deserializing multiple objects
I am deserializing multiple objects from a slice (like in https://github.com/pyfisch/cbor/issues/20).

I am using the EOF error to detect when I reached the last object, but I have no way to distinguish between an actual EOF and an invalid cbor object.

It would be great to have an `Iter` implementation that does the right thing or that the thrown error had not the current offset in the stream but the offset where the deserialization began.

sfackler:
Closes #68

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -6,6 +6,7 @@ use std::io;
 use std::str;
 use std::f32;
 use std::result;
+use std::marker::PhantomData;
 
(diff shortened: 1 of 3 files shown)


pyfisch:
Please rebase it.

sfackler:
Updated!

PR #75, opened 2018-10-06
Don't pass Visitor to parse_indefinite_{bytes,str}

bytwise:
It was used to create a `V::Value`. Because that was the last thing the
methods did, we can move the creation to the call sites instead.
Before this change, Rustc would compile a new version of these methods
for each `Visitor`.

This should improve compile times of crates using `serde_cbor`.

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -191,10 +191,7 @@ where
         }
     }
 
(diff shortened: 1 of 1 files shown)


sfackler:
Thanks!

PR #81, opened 2018-11-20
[WIP] Move scratch buffers into readers

chrysn:
This PR describes an idea of how to one might continue after #80, and one answer to the question of whether `scratch_offset` can go away there.

The meat of this PR is 86a51c3762d7daabc214f3621e0d63ed8cd91b4f, 18e254623a39c0c4b0fb69f13181a18f37c9ea70 and 9e123f7cdb4f9492e922c6af72fe314dc3f7b148 (there's a few before because I can't create a PR in github that's based on another PR) which may also make sense to read as a cumulative diff; between them, there is 6fa55802f6a1ae7ce34291ce6aa107a4d6981476 that describes a possible application, and then interfacing and test coverage.

What the patch does is to move the scratch buffer from the deserializer to the the Read trait; each reader can manage a buffer as it seems fit. The Read trait methods change a bit:
* read() now always returns a slice, but while earlier it gave an Enum telling you to either take the 'de long-lived result or to have a short-lived peek into the buffer, it now returns an Enum that always contains a slice but has variable lifetimes.
* For gathering a long slice, the's now the `clear_buffer()` / `read_to_buffer()` / `view_buffer` trio, which exposes functionality previously managed by the decoder's buf (`buf.clear()`, `.read(n, buf)`, `buf.as_ref()`) -- that allows implementing MutSliceRead to manage its own scratch buffer by shuffling data around in its slice. For consistency, `.view_buffer()` returns the same enum as `.read()`. (This allows indefinite strings from a MutSliceRead to go the `visit_borrowed_bytes` route).

The rationale for these changes is twofold:
* They (or at least the first part thereof in 86a51c3762d7daabc214f3621e0d63ed8cd91b4f) allow implementing a MutSliceRead, and thus decoding any CBOR in a no_std (and not even alloc) environment.
* They reduce the `#[cfg(...)]` gating that'd be necessary to make serde_cbor ready for no_std because the individual readers now can take the cfg decisions. (And even there it's fewer places, as the Read trait can stay the same, IoRead needs to be gated anyway, and only SliceRead needs to decide whether it can do `.read_to_buffer()` or not.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: rust
 rust:
-  - 1.18.0
+  - 1.26.0
   - stable
   - beta
   - nightly
--- a/src/de.rs
+++ b/src/de.rs
@@ -11,7 +11,7 @@ use std::marker::PhantomData;
 
 use error::{Error, Result, ErrorCode};
 use read::EitherLifetime;
-pub use read::{Read, IoRead, SliceRead};
+pub use read::{Read, IoRead, SliceRead, MutSliceRead};
 
 /// Decodes a value from CBOR data in a slice.
 ///
@@ -39,6 +39,21 @@ where
     T: de::Deserialize<'a>,
 {
     let mut deserializer = Deserializer::from_slice(slice);
+    let value = de::Deserialize::deserialize(&mut deserializer)?;
+    deserializer.end()?;
+    Ok(value)
+}
+
+/// Decode a value from CBOR data in a mutable slice.
+///
+/// This can be used in analogy to `from_slice`. Unlike `from_slice`, this will use the slice's
+/// mutability to rearrange data in it in order to resolve indefinite byte or text strings without
+/// resorting to allocations.
+pub fn from_mut_slice<'a, T>(slice: &'a mut [u8]) -> Result<T>
+where
+    T: de::Deserialize<'a>,
+{
+    let mut deserializer = Deserializer::from_mut_slice(slice);
     let value = de::Deserialize::deserialize(&mut deserializer)?;
     deserializer.end()?;
     Ok(value)
@@ -98,6 +113,16 @@ impl<'a> Deserializer<SliceRead<'a>> {
     /// Borrowed strings and byte slices will be provided when possible.
     pub fn from_slice(bytes: &'a [u8]) -> Deserializer<SliceRead<'a>> {
         Deserializer::new(SliceRead::new(bytes))
+    }
+}
+
+impl<'a> Deserializer<MutSliceRead<'a>> {
+    /// Constructs a `Deserializer` which reads from a mutable slice that doubles as its own
+    /// scratch buffer.
+    ///
+    /// Borrowed strings and byte slices will be provided even for indefinite strings.
+    pub fn from_mut_slice(bytes: &'a mut [u8]) -> Deserializer<MutSliceRead<'a>> {
+        Deserializer::new(MutSliceRead::new(bytes))
     }
 }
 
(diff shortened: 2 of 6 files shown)


pyfisch:
Thanks.

Can you rebase it on master. I think will make review easier.

chrysn:
Rebased and slightly simplified (in terms of back and forth between function names).

I'd still call this PR "work in progress" because the MutSliceRead implementation is added in the middle of them (which makes sense from a rationale and story perspective), but I'm not sure whether you're OK with having unsafe parts in this library. Other than that, I think it's good for review. (And if you conceptually like the changes but want MutSliceRead to wait for a dedicated unsafe-wrapping crate, I'll happily rebase it out.)

pyfisch:
> error[E0433]: failed to resolve: use of undeclared type or module `Reference`
   --> src/read.rs:138:23
    |
138 |             return Ok(Reference::Borrowed(&[]));
    |                       ^^^^^^^^^ use of undeclared type or module `Reference`

chrysn:
I should have known better than to rebase and not to check the ensuing trivial changes.

The error is now fixed by a better rebase, and I've incorporated two other small fixes in the patch set right away (`view_buffer` had a needlessly mutable self, and turning the `conver_str` into an associated function was insufficiently motivated).

As for the the `convert_str` rationale, I might add here: Having `read` return `EitherLifetime` is IMO safer w/rt the usage pattern -- it doesn't demand of the caller to follow the `clear` / `read_to_buffer` / `view_buffer` sequence that a caller of `read_to_buffer` needs to follow, and allows for much consistency between `view_buffer` and `read`. The downside of this is that a reader is mutably borrowed while a `read`'s results are being processed, necessitating the `convert_str` to become an associated method. If you'd rather have it another way (with `.read()` returning something more like the old `Reference` and a dedicated `.view_buffer()` if it's a copy), I can rearrange that -- my opinion is that the way in the PR is the slightly clearer one.

chrysn:
Addressing the travis failure: Is 1.18 the earliest version that's targeted for support? Then I go check in detail whether it's anything else than the MutSliceRead that fails with those versions.

pyfisch:
> Addressing the travis failure: Is 1.18 the earliest version that's targeted for support?

You can update the minimum version to 1.30. It is mainly there so I am notified of changes for old versions of rust.

chrysn:
Sorry I missed that the ball was in my court since the last update. Added f1884a95b4e2776b15eb2184dfaf083cb2fd8b3a to bump the "old" version travis tests with; checks pass now.

chrysn:
The last commit added atop this changes MutSliceRead to use the [SealingSlice](https://crates.io/crates/sealingslice) abstraction created with this application in mind. I'm fully open on whether or not you'd like to go that route (having an effectively 12-lines-of-code (but much more to test it) dependency rather than having a single "unsafe" in the own code), I would.

chrysn:
I'm retracting the last change that'd work on SealingSlice; a [review on that crate](https://gitlab.com/chrysn/sealingslice/issues/1) pointed me towards a way to implement what we need here (not all of SealingSlice, but that overreached) in a safe way (ie. without bespoke unsafe code either in this crate or sealingslice). This changes the internal interface introduced in this PR again slightly, but was left as commits atop the patch series.

If you haven't started reviewing yet, I can squash that API change back into sequence, but so far left the two top-most commits (89cef504ef9d81fa56ca70d3fc663e56e0763da0 and 964c454bb672d6502a520cbea15efafadbc487c8) as individual ones to be better readable as a patch. When those are squashed, `.view_buffer()` will be renamed to `.take_buffer()` or similar to reflect the slightly changed semantics.

pyfisch:
Thanks a lot :tada:

PR #82, opened 2018-11-28
sort ObjectKey using canonical ordering

kcking:
This should check an item off of https://github.com/pyfisch/cbor/issues/17. Let me know if you have any questions or concerns! I took care to implement this without needing to serialize the keys


src/value/value.rs
```diff
@@ -374,7 +374,53 @@ pub enum ObjectKey {
...
+    fn partial_cmp(&self, other: &ObjectKey) -> Option<std::cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
 impl ObjectKey {
+    // Returns a key to sort `ObjectKey`s by major type, then byte-length, then lexicographically
+    // by content. Equivalent to sorting `ObjectKey`s __lexicographically__ by their CBOR serialization.
+    //
+    // Note: the first guidelines regarding canonicalized CBOR
```
pyfisch: Why was the canonical order changed?

pyfisch:
I see this orders only ObjectKey but also lists and even objects can be used as keys. 

With this change it should be relatively easy to serialize `Value` in canonical order but not arbitrary structures with derive.

kcking:
Thanks for taking a look at the changes.

Ah yes this doesn't affect derived struct or map serializations. After looking further, it looks like this might be a limitation of serde and the order of serialization of map keys and struct fields is out of our control.

I suppose my next step would be to look into adding that to serde.

pyfisch:
I am not sure that serde would accept this functionality. We should try before to make it work in serde-cbor even if it is more limited and has a worse API. It should be possible to serialize structs and enums to a CBOR-Value and then canonically serialize this value. (De-serialization is in reverse.) This will be slower and require temporary allocations for the whole message but may be doable without help from serde.

src/value/value.rs
kcking: Some of the discussion regarding the change is located [here](https://www.ietf.org/mail-archive/web/cbor/current/msg00264.html), but mainly for implementation simplicity in that values of different types do not need to be compared.

kcking:
I found a way to achieve map sorting in the serializer impl by overriding `collect_map`. I also modified struct serialization to not write anything until `.end()` is called and do the sorting then.

I haven't given any thought to deserialization -- I suppose a `from_slice_canon` method could re-serialize the value and check that it matches the passed-in bytes. Another method would be to check in real-time that keys are sorted and ints are smallest encoding.

pyfisch:
:tada:

sfackler:
Am I missing something or is there no way to opt out of buffering the entire input and sorting it?

kcking:
Correct, would be easy to add though! Maybe canonicalization should be opt-in

Issue #79, opened 2018-11-12
no_std support


chrysn:
Serde supports no_std nicely (judging from the documentation); it would be nice if serde_cbor could too.

I'm currently looking around for what parts of the crate would need to be conditional on std's presence, so basically whether cfg(feature=...) gating; will keep this issue updated.

chrysn:
The single thing that makes this tricky is the prevalence of IO errors throughout the return codes -- many things possibly return an std::io::Error which is unknown to no_std environments, even though they have no chance of actually producing one. Looking at different variants of how to solve this -- my current string replacement orgy of annotating every error or result with its IO error type (for lack of higher-kinded types) probably won't be the solution, but might should be helpful in finding out what else is needed.

sfackler:
I was going to suggest looking at serde_json, but it looks like that doesn't have no_std support yet!

chrysn:
> serde_json, but it looks like that doesn't have no_std support yet

There is serde_json_core, but that seems to be completely independent --
which is IMO not something to strive for.

I think that with some flexibility in the error handling (eg. an error
type that can be IOError, and one that can't but is
Into<ErrorMaybeEvenIO>), serde_cbor can have its "core" parts in it
unconditionally (along with the simpler impls for `de` that work off
slices and that like), and if std is turned on (as is common to do by
default), also impl for io::{Read,Write}.

What I'm stil trying to figure out is with how little upheaval in terms
of public signatures that can happen, and the open question to the room
is how much such upheaval the project wants to tolerate in order to
achieve no_std compatibility.

dtolnay:
For serde_json we intend to provide a no_std mode that requires alloc. The no alloc case is covered by serde_json_core.

[`serde_json::ser::Formatter`](https://docs.rs/serde_json/1.0.33/serde_json/ser/trait.Formatter.html) uses `io::Error` but in no_std mode we can make all those methods private and disallow downstream impls of the trait. Haven't looked at whether it is a more complicated situation in this crate.

chrysn:
The no_std use I'm aiming for is more than just not having std in public
types (as using private to prevent impls would achieve) -- I'm working
with what effectively is Rust on bare metal on microcontrollers, and
there is just no std::io defined (there is no concept of an operating
system's file there either).

The use of alloc is two-fold as far as I've read the code: The Value type
obviously depends on it (being a recursive data structure), but that
can be cfg-gated. The larger issue is with the Error type carrying
[...]

PR #89, opened 2019-01-22
no_std: more preparatory steps

chrysn:
While work on #79 was set back by the canonicalization changes (which introduced quite some internal allocation), I'd like to propose a few next steps here for inclusion in the next-version branch.

* 39125f4f0335ac5502db8778f61c4ca8acda1df5 is a renaming step after #81 (originally proposed to happen after review was through to make it easier to follow through the code changes in https://github.com/pyfisch/cbor/pull/81#issuecomment-447261360 -- anyway, should be done as the current name is slightly misleading).
* 0818e2f719b5e9f8e386763a805394b668716d5d just removes a level of needless heap-allocation that's contained within the error module
* 690483435d76ab3d5c77359b638f7c09be6d1cec switches from using std to core wherever feasible (became less of an issue in 2018 edition with the `extern crate core;` being gone); doesn't change the built files because anything used from std here is just a re-export from core
* e4619bcda4bfc5113c963774bc9aca7dfe1d4848 moves the custom IO errors the readers use out of the read signature, and consistently uses error::Result instead.

### DIFF
--- a/examples/readme.rs
+++ b/examples/readme.rs
@@ -1,8 +1,8 @@
 // NOTE: This file should be kept in sync with README.md
 
+use serde_derive::{Deserialize, Serialize};
 use std::error::Error;
 use std::fs::File;
-use serde_derive::{Deserialize, Serialize};
 
 // Types annotated with `Serialize` can be stored as CBOR.
 // To be able to load them again add `Deserialize`.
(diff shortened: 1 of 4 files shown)


pyfisch:
Hi, sorry for the delay. I hope I can find time to review this on the weekend.

pyfisch:
> While work on #79 was set back by the canonicalization changes

I have changed canonicalization again. All CBOR values will be serialized in canonical order. Probably values won't be supported in no std mode (or be limited).

PR #95, opened 2019-02-27
Implement Error::source when available

### LINKED ISSUE #94: Retrieve underlying error from error::Error
I would like to get the e.g. IO-Error from the Error type. So it would be senseful to implement source() or a custom method which can achieve this.

sfackler:
Closes #94

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,9 @@ byteorder = "1.0.0"
 half = "1.2.0"
 serde = "1.0.14"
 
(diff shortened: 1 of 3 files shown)


pyfisch:
I am working on a new version on the next-version branch. It will support only rust 1.31+. Can you please create a PR for this branch and without the rustc_version crate?

sfackler:
Done! https://github.com/pyfisch/cbor/pull/97

PR #97, opened 2019-03-02
Update Error impl

### LINKED ISSUE #94: Retrieve underlying error from error::Error

sfackler:
Remove the implementation of the now-deprecated description method, and
implement source rather than cause.

Closes #94


pyfisch:
Thanks :+1:

PR #100, opened 2019-03-16
Support deserializing from mixed slices

baloo:
When trying to parse fido2 (ctap2) responses from token,
slices are packed together, cbor encoded data may be followed
by other type of data.

This patch introduces a new `from_slice_stream` method that would
return the rest of the buffer instead of raising an error.

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -44,6 +44,34 @@ where
     Ok(value)
 }
 
+/// Decodes a value from CBOR data in a longer than necessary slice.
+///
+/// This is used when mixed data are packed together
+///
+/// # Examples
+///
+/// Deserialize a `String`
+///
+/// ```
+/// # use serde_cbor::de;
+/// let v: Vec<u8> = vec![0x66, 0x66, 0x6f, 0x6f, 0x62, 0x61, 0x72,
+///                       0x66, 0x66, 0x6f, 0x6f, 0x62, 0x61, 0x72];
+/// let (rest, value): (&[u8], String) = de::from_slice_stream(&v[..]).unwrap();
+/// assert_eq!(value, "foobar");
+/// assert_eq!(rest, &[0x66, 0x66, 0x6f, 0x6f, 0x62, 0x61, 0x72]);
+/// let (rest, value): (&[u8], String) = de::from_slice_stream(rest).unwrap();
+/// assert_eq!(value, "foobar");
+/// assert_eq!(rest, &[]);
+/// ```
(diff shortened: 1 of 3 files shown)


baloo:
`from_slice_stream` does not sound right, but I couldn't come up with a better name. Let me know if you have one.

dtolnay:
As an alternative we could expose the current byte offset of Deserializer\<SliceRead\>, similar to
https://docs.rs/serde_json/1.0.39/serde_json/de/struct.StreamDeserializer.html#method.byte_offset

baloo:
That could work too, but we shouldn't throw an error I think.

dtolnay:
Which error are you referring to? Serde_json's `byte_offset()` does not return error.

baloo:
``` rust
    use serde_cbor::{from_slice, from_slice_stream};
    #[test]
    fn parse_reader_slice() {
        let v: Vec<u8> = vec![0x66, 0x66, 0x6f, 0x6f, 0x62, 0x61, 0x72, 0x66, 0x66, 0x6f, 0x6f, 0x62, 0x61, 0x72];
        let value = from_slice(&v[..]);
        let value: String = value.unwrap();
        assert_eq!(value, "foobar");
    }   

    #[test]
    fn parse_reader() {
        let v: Vec<u8> = vec![0x66, 0x66, 0x6f, 0x6f, 0x62, 0x61, 0x72, 0x66, 0x66, 0x6f, 0x6f, 0x62, 0x61, 0x72];
        let (rest, value): (&[u8], String) = from_slice_stream(&v[..]).unwrap();
        assert_eq!(value, "foobar");
        assert_eq!(rest, &[0x66, 0x66, 0x6f, 0x6f, 0x62, 0x61, 0x72]);
        let (rest, value): (&[u8], String) = from_slice_stream(rest).unwrap();
        assert_eq!(value, "foobar");
        assert_eq!(rest, &[]);
    }   
```

from_slice in the first example would return:
```
Err(ErrorImpl { code: TrailingData, offset: 8 })
```

Altough the error contains the offset, and I could workaround this, I still thinks returning the rest of the slice is more natural.

dtolnay:
If you are suggesting that from_slice be changed to silently ignore trailing garbage, I don't agree with that. The caller would write (untested):

```rust
let mut deserializer = serde_cbor::Deserializer::from_slice(slice);
let value = T::deserialize(&mut deserializer)?;
let rest = &slice[deserializer.byte_offset()..];
```

baloo:
V2 in #101

PR #101, opened 2019-03-18
Support deserializing from mixed slices.

baloo:
When trying to parse fido2 (ctap2) responses from token,
slices are packed together, cbor encoded data may be followed
by other type of data.

This patch introduces a new `byte_offset` method that would
return the position of the reader.

Based on suggestions of @dtolnay. Superseed #100.

### DIFF
--- a/tests/de.rs
+++ b/tests/de.rs
@@ -48,6 +48,7 @@ mod std_tests {
     use serde_bytes::ByteBuf;
     use std::collections::BTreeMap;
 
+    use serde::de as serde_de;
     use serde_cbor::{de, error, from_reader, to_vec, Deserializer, ObjectKey, Value};
 
     #[test]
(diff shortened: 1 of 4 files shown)


pyfisch:
LGTM. Can you please rebase the changes and add a simple test?

baloo:
@pyfisch rebased and tested

baloo:
@pyfisch any chance you could merge this?

tests/de.rs
```diff
@@ -1,3 +1,4 @@
+extern crate serde;
```
pyfisch: This `extern crate` statement shouldn't be needed with the 2018 edition.

pyfisch:
LGTM. Will merge with the comment addressed.

If you have time it would be great if you could add an example to the crate level documentation in `lib.rs` (right before the section "limitations")

baloo:
@pyfisch done

pyfisch:
@baloo thanks

PR #103, opened 2019-03-18
Implement ser/de for 128 bit numbers

pyfisch:
Serde can represent negative numbers up
to -2^64 while an i64's minimum value
is -2^63-1. For this reason these
numbers are now stored as i128.

For symmetry u128 is added.
Note that no numbers beyond +/- 2^64
can be expressed in CBOR and will raise
an error.


pyfisch:
Additionally this prepares serde-cbor for an improved `Value` type.

PR #105, opened 2019-03-24
Split internal SerializeStruct into itself

pickfire:
`SerializeStruct` does not seem to require next_key.

I based this off https://github.com/dtolnay/serde-yaml/blob/93f4d40d04ad2c424a58ba1fa34f55a3d0faab4f/src/ser.rs#L33 while working on https://github.com/dtolnay/serde-yaml/issues/44, maybe this could be applied to serde-json?

CC @dtolnay

### DIFF
--- a/src/value/ser.rs
+++ b/src/value/ser.rs
@@ -27,7 +27,7 @@ impl serde::Serializer for Serializer {
     type SerializeTupleStruct = SerializeVec;
     type SerializeTupleVariant = SerializeTupleVariant;
     type SerializeMap = SerializeMap;
-    type SerializeStruct = SerializeMap;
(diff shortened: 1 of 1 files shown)


sfackler:
Could you clarify what problem this is solving?

pickfire:
@sfackler The struct will have not have `next_key`, it does not seemed to be used.

sfackler:
Why is that a problem?

pickfire:
@sfackler Wouldn't it be better to remove a variable (garbage) that is not used? It helps to save memory.

Are there any issues with removing a field that is not used?

pyfisch:
Thanks, but I don't think this optimization is needed. I will change `Value` soon anyway so after that I may look into it again.

PR #106, opened 2019-03-30
fixup serde_bytes dependency

baloo:
When a crate depends on serde_cbor master, the following error is
thrown:
```
error: failed to select a version for `serde_cbor`.
    ... required by package `authenticator v0.2.4 (/home/baloo/work/dev/authenticator-rs)`
versions that meet the requirements `*` are: 0.9.0

the package `authenticator` depends on `serde_cbor`, with features: `serde_bytes` but `serde_cbor` does not have these features.

failed to select a version for `serde_cbor` which could resolve this conflict
```

I believe serde_bytes should not be a dev-dependency but an optional
dependency instead.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,9 +20,9 @@ maintenance = { status = "actively-developed" }
 byteorder = { version = "1.0.0", default-features = false }
 half = "1.2.0"
 serde = { version = "1.0.14", default-features = false }
+serde_bytes = { version = "0.10", default-features = false, optional = true }
 
 [dev-dependencies]
-serde_bytes = { version = "0.10", default-features = false }
 serde_derive = { version = "1.0.14", default-features = false }
 
 [features]


pyfisch:
But why should it be an optional dependency, it is only used in the tests?

baloo:
this looks to be a feature missing/bug in cargo that looks for `serde_bytes` dependency because it is mentioned in the `features` section (`std`)

baloo:
Although `serde_bytes` is only required for documentation (as far as I understand it), my understanding of the documentation is that they should be specified as optional dependency:
https://doc.rust-lang.org/cargo/reference/manifest.html#the-features-section

But I'm not sure I'm reading it correctly.

pyfisch:
While this fixes the immediate problem `serde_cbor` with the `std` feature enabled now *always* builds `serde_bytes` too.

```
$ cargo build
   Compiling serde v1.0.89
   Compiling byteorder v1.3.1
   Compiling half v1.3.0
   Compiling serde_bytes v0.10.5
   Compiling serde_cbor v0.9.0 (/home/michael/Dokumente/git/cbor)
    Finished dev [unoptimized + debuginfo] target(s) in 17.96s
```

baloo:
@pyfisch let me know if you want me to submit another PR with another approach, but I would really need to see master build :(.

baloo:
fixed in #114

Issue #110, opened 2019-05-30
Add byte_offset() method for StreamDeserializer.


ChaseElectr:
I'm using serde_cbor as serializer and store datas in a file side by side. When I read the file and deserializing datas, I want to remember every data's offset in the file so that I don't need to keep the entire file in the memory. But only the struct Deserializer has a byte_offset() method, and StreamDeserializer not. I think we can easily add it.

pyfisch:
I believe this is done?

PR #111, opened 2019-05-30
Add byte_offset() method for StreamDeserializer.

ChaseElectr:
When iterating over the input stream during deserializtion, we have no way to get the data size in the stream. This function can help us a little about that.
 #110

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -1140,6 +1140,18 @@ where
     }
 }
 
+impl<'de, R, T> StreamDeserializer<'de, R, T>
+where
+    R: Offset,
+    T: de::Deserialize<'de>,
+{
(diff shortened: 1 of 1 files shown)


ChaseElectr:
I do have thought it, but the Deserializer doesn't implement Offset:
https://github.com/pyfisch/cbor/blob/9df0c6f0ab8af0d0782c085318caea33df626bc7/src/de.rs#L800
I didn't see the reason for it, just thought I might not fully understand the whole picture :p

pyfisch:
>  I didn't see the reason for it, just thought I might not fully understand the whole picture :p

Apparently me neither. :p
`Offset` is just used for readers and de-serializers just have a method `byte_offset` if possible.

PR #114, opened 2019-06-06
Remove serde-bytes dev-dependency

pyfisch:
It caused build problems because of optional features.
Change some tests for byte strings.
Remove failing tests I believe to be redundant.

cc @baloo


baloo:
@pyfisch many thanks!

baloo:
I can confirm that fixes the problems for me.

Issue #109, opened 2019-05-12
serialize_f64 is too clever


Nadrieril:
The [`serialize_f64`](https://github.com/pyfisch/cbor/blob/9df0c6f0ab8af0d0782c085318caea33df626bc7/src/ser.rs#L471-L479) method of the cbor serializer tries to be clever: if the provided value fits in a `f32` or `f16`, it uses the shorter representation to save bytes.
I'm using cbor in a context where the output is passed through a hash function, so the precise representation chosen is important. The problem I have is that I have no way of telling the cbor serializer to use only the `f64` size.
Would it be possible to make the `serialize_f64` method always use the `f64` representation ? I also think that's what would be expected by serde consumers in general.

pyfisch:
[Section 3.6](https://tools.ietf.org/html/rfc7049#section-3.6) states that an encoder can encode a number in any way it likes to as long as the numeric value stays identical but applications may specify something else.

It should be noted that serde can't decode integer values if they are stored as floats (If I recall correctly).

> I'm using cbor in a context where the output is passed through a hash function, so the precise representation chosen is important.

As long as the encoder always uses the same encoding scheme I don't see a problem with using `f32` and `f16` in place of `f64`.

> Would it be possible to make the serialize_f64 method always use the f64 representation ? I also think that's what would be expected by serde consumers in general.

Can you give an example of an encoder expecting this?

Nadrieril:
> As long as the encoder always uses the same encoding scheme I don't see a problem with using `f32` and `f16` in place of `f64`.

That's only if there is only one encoder used. I'm working on the Rust implementation of the [Dhall language](https://github.com/dhall-lang/dhall-lang), and this language in particular allows adding integrity checks to files. Those integrity checks work by encoding the file as binary (following a precise encoding specification), and then hashing the result. But because the language has several implementations (in different languages), it is important that all of them use the exact same binary encoding.
The dhall standard specifies how to encode doubles [here](https://github.com/dhall-lang/dhall-lang/blob/master/standard/binary.md#double). I don't see a way to use `serde_cbor` to produce that encoding. I tried using `serialize_bytes` but that can't work to produce floats.

Nadrieril:
Here's my reasoning behind changing the behavior. There are two possible behaviors one may want when encoding a f64 value:
1. encode the f64 as the smallest representation possible;
2. encode the f64 as f64.

[...]

PR #115, opened 2019-06-08
Update re-exports and improve value

pyfisch:
Only export the [recommended items](https://serde.rs/conventions.html) from the crate root. Are there enough `to_*` and `from_*` functions exported? I think it is ok that the functions for no-std are found in the `ser` and `de` modules.

Rename all the variant names of the CBOR value to reflect the naming from the RFC more closely. Remove the distinction between I64 and U64. Always store values in canonical order. (serializing CBOR in canonical form should work now :tada: ) I removed the value test and access methods like `is_object` and `get_object`. Are they used in real projects? If this is the case I can add them back in.

What do you think?


sfackler:
LGTM

Might be worth removing the forced canonicalization in the raw streaming serializer in favor of having people serialize into a Value and then write that out instead?

pyfisch:
> Might be worth removing the forced canonicalization in the raw streaming serializer in favor of having people serialize into a Value and then write that out instead?

Yes, the idea is that people who need canonical CBOR documents serialize their data into `Value` and then this is serialized as canonical CBOR with the correct order of keys inside maps.
Which parts of canonicalization would you remove from the streaming serializer? (see also #109)

sfackler:
The part where the struct serializer buffers all of its entries in memory: https://github.com/pyfisch/cbor/blob/master/src/ser.rs#L771

pyfisch:
>The part where the struct serializer buffers all of its entries in memory: https://github.com/pyfisch/cbor/blob/master/src/ser.rs#L771

Oh, I forgot about this. Yes I will remove the buffering.

Issue #117, opened 2019-06-09
Configuration for Serializer and Deserializer


pyfisch:
Idea:
```rust
pub struct SerializerConfig {
    packed: bool,
    self_describe: bool,
    legacy_enums: bool,
}

pub struct DeserializerConfig {
    accept_named: bool,
    accept_packed: bool,
    accept_standard_enums: bool,
    accept_legacy_enums: bool,
}

impl Default for SerializerConfig {
    fn default() -> Self {
        Self {
            packed: false,
            self_describe: false,
            legacy_enums: false,
        }
    }
}

impl Default for DeserializerConfig {
    fn default() -> Self {
        Self {
            accept_named: true,
            accept_packed: true,
            accept_standard_enums: true,
            accept_legacy_enums: true,
        }
    }
}

impl From<SerializerConfig> for DeserializerConfig {
    fn from(config: SerializerConfig) -> Self {
        Self {
            accept_named: !config.packed,
            accept_packed: config.packed,
            accept_standard_enums: !config.legacy_enums,
            accept_legacy_enums: config.legacy_enums,
        }
    }
}
```

Which options are missing?

pyfisch:
Options are implemented in v0.10.0

PR #118, opened 2019-06-09
Serialize enums as maps by default

### LINKED ISSUE #85: Change enum serialization
https://serde.rs/enum-representations.html

Serde CBOR should use the "Externally tagged" format by default in the next breaking version.
Also make sure that the other formats round-trip.
The current array based serialization should be kept as a legacy format so old data can be read.

pyfisch:
This matches other serde formats.
Directly set options on Serializer.
Remove rarely used to_* methods.

Closes #85

### DIFF
--- a/src/ser.rs
+++ b/src/ser.rs
@@ -11,58 +11,6 @@ use serde::ser::{self, Serialize};
 #[cfg(feature = "std")]
 use std::io;
 
-/// Serializes a value to a writer.
-#[cfg(feature = "std")]
-pub fn to_writer<W, T>(writer: &mut W, value: &T) -> Result<()>
-where
-    W: io::Write,
-    T: ser::Serialize,
-{
-    value.serialize(&mut Serializer::new(&mut IoWrite::new(writer)))
-}
-
-/// Serializes a value to a writer and adds a CBOR self-describe tag.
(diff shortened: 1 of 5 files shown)


pyfisch:
Search results reveal that [to_vec], [to_writer] and [to_vec_packed] are primarily used. Other methods such as [to_vec_sd] are only rarely needed. Therefore users should directly call the Serializer in this case.

[to_vec]: https://github.com/search?l=Rust&q=serde_cbor%3A%3Ato_vec&type=Code
[to_writer]: https://github.com/search?l=Rust&q=serde_cbor%3A%3Ato_writer&type=Code
[to_vec_packed]: https://github.com/search?l=Rust&q=serde_cbor%3A%3Ato_vec_packed&type=Code
[to_vec_sd]: https://github.com/search?l=Rust&q=serde_cbor%3A%3Ato_vec_sd&type=Code

pyfisch:
Regarding #117, I felt that a config struct without accessors may hinder future extensibility and a separate builder was too much complexity.

It is now possible in theory so change the options in the middle of serialization and shoot yourself in the foot but this requires some effort on the users side.

sfackler:
Yeah, I think just having setters on Serializer/Deserializer for options makes sense.

Issue #119, opened 2019-06-10
Release


baloo:
The last release available for this crate is 0.9 which is 10 month old. There is a couple of improvement pending in master that I would need, and a newer release would be nice.

@pyfisch It looks like you've been submitted a couple of changes lately. Would you be able to list all tasks/fixes required before a release? I can try to help with some of them.

Thanks!

pyfisch:
Hi @baloo, I would appreciate your help.

These are the things that should be done before a release:

* [ ] complete the [deserializer-options](https://github.com/pyfisch/cbor/tree/deserializer-options) branch
  * [ ] test that if `accept_standard_enums` is false no enums-as-map can be deserialized
  * [ ] test that if `accept_legacy_enums` is false no enums formatted as arrays can be deserialized
  * [ ] if both are false no enums at all can be deserialized
  * [ ] (optional) find a way to disable the packed format in the deserializer (See #112), but I only want to support this if it does not require much extra code
    * [ ] write tests for this
* [ ] write a How-To for no-std usage (should be included on main page of API docs)
* [ ] add an example to the `StreamDeserializer`
* [ ] update main page of API docs (I can do this myself, but if you want to do this, reach out to me and ask me what should be in the text)

baloo:
I have a hard time understanding how the packed format is actually deserialized. I'm not sure I can pull that one off :(

pyfisch:
> I have a hard time understanding how the packed format is actually deserialized. I'm not sure I can pull that one off :(

Yeah, I initially assumed it was implemented in this crate but mapping the numbers to named fields and variants appears to be built-in serde magic.

baloo:
@pyfisch let me know if you want me to do the last item on the list, and as mentioned, if so let me know about any content you'd want on that page.

pyfisch:
@baloo for the docs just have a look that they are up-to-date (e.g. don't reference any removed functions. The long example document with John Doe should be removed as I don't think anymore it is that helpful. Instead there should be a usage example like in the readme (or the same example copied).

baloo:
Anything missing for a release?

baloo:
@pyfisch up?

pyfisch:
## Making a Release

* [ ] Make sure the crate compiles and all tests pass.
* [ ] (Optional) Test that the fuzzer works and fuzz the crate for some time.
* [ ] Write a list with all changes made since the last release
[...]

PR #121, opened 2019-06-12
Test deserializer options

baloo:
Test format generated by 0.9 and 0.10 for enum types and pass try them through `accept_standard_enums` and `accept_legacy_enums` options.

See #119 for details

### DIFF
--- a/tests/de.rs
+++ b/tests/de.rs
@@ -402,4 +402,84 @@ mod std_tests {
         assert_eq!(value, "foobar");
         assert_eq!(rest, &[]);
     }
+
+    fn from_slice_stream_options<'a, T>(
+        slice: &'a [u8],
+        standard: bool,
+        legacy: bool,
+    ) -> error::Result<(&'a [u8], T)>
+    where
+        T: serde_de::Deserialize<'a>,
+    {
+        let deserializer = Deserializer::from_slice(slice);
+        let deserializer = if !standard {
+            deserializer.disable_standard_enums()
+        } else {
+            deserializer
+        };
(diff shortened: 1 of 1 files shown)


pyfisch:
LGTM, can you check that the "Enum::Unit" variant is correctly deserialized if one option is enabled but not otherwise?

baloo:
I'm not sure what you mean here,  because `Enum::Unit` is expected to be serialized the same way in both version 0.9 and 0.10:
```
    /// ## Default encodings
    /// 
    /// * `Enum::Unit` encodes as `"Unit"`
    /// * `Enum::NewType(10)` encodes as `{"NewType": 10}`
    /// * `Enum::Tuple("x", true)` encodes as `{"Tuple": ["x", true]}`
    /// 
    /// ## Legacy encodings
    /// 
    /// * `Enum::Unit` encodes as `"Unit"`
    /// * `Enum::NewType(10)` encodes as `["NewType", 10]`
    /// * `Enum::Tuple("x", true)` encodes as `["Tuple", "x", true]`
    /// * `Enum::Struct{ x: 5, y: -5 }` encodes as `["Struct", {"x": 5, "y": -5}]`
```

I added a test for `Enum::Unit` to confirm.

pyfisch:
> I'm not sure what you mean here, because Enum::Unit is expected to be serialized the same way

I want to be sure it can't be deserialized if both formats are disabled.
https://github.com/pyfisch/cbor/pull/121/files#diff-eaeb4d5a98ecdfb3e95af5b2a5867a25R474

baloo:
Alright! Thanks!

PR #122, opened 2019-06-12
Adds an example to StreamDeserializer

baloo:
see #119 

@pyfisch I wasn't quite sure what you needed as example, this looks explicit enough to me, but let me know if you have any other ideas.


pyfisch:
That should help! Thanks a lot.

PR #123, opened 2019-06-12
no-std: Fixup warning on dead code

baloo:
When building cbor with no-std, I get the following warning:
```
warning: variant is never constructed: `Io`
   --> cbor/src/error.rs:271:5
    |
271 |     Io,
    |     ^^
    |
    = note: #[warn(dead_code)] on by default
```


baloo:
I guess this is more a question than a pull request, I'm not entirely sure why that was here, but if a dependent crate do not use std, they should not expect Io error either. Or was that on purpose to not break api compability too much between std/no-std variants?

pyfisch:
Let us remove it and add it back when needed.

PR #124, opened 2019-06-12
Document no-std support

baloo:
see #119

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -206,20 +206,32 @@
 //!     password_hash: [u8; 4],
 //! }
 //!
-//! let mut value = [
+//! let value = [
 //!     0xa2, 0x67, 0x75, 0x73, 0x65, 0x72, 0x5f, 0x69, 0x64, 0x18, 0x2a, 0x6d,
 //!     0x70, 0x61, 0x73, 0x73, 0x77, 0x6f, 0x72, 0x64, 0x5f, 0x68, 0x61, 0x73,
 //!     0x68, 0x84, 0x1, 0x2, 0x3, 0x4
 //! ];
-//! let mut scratch = [0u8; 32];
 //!
+//! // from_slice_with_scratch will not alter input data, use it whenever you
+//! // borrow from somewhere else.
+//! // You will have to size your scratch according to the input data you
+//! // expect.
 //! use serde_cbor::de::from_slice_with_scratch;
+//! let mut scratch = [0u8; 32];
 //! let user: User = from_slice_with_scratch(&value[..], &mut scratch)?;
 //! assert_eq!(user, User {
 //!     user_id: 42,
 //!     password_hash: [1, 2, 3, 4],
 //! });
 //!
+//! let mut value = [
+//!     0xa2, 0x67, 0x75, 0x73, 0x65, 0x72, 0x5f, 0x69, 0x64, 0x18, 0x2a, 0x6d,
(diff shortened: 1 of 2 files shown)


pyfisch:
Can you move the no-std section before the limitations?

What to write in the Cargo.toml is important but I'd like to see an example (or explanation) how to serialize and deserialize a simple struct without std. (How to use `SliceWrite` and `SliceRead`, `from_mut_slice`, `from_slice_with_scratch`, btw. the last two methods should be visible in the docs also when `std` is enabled for better discoverability)

baloo:
@pyfisch there

src/lib.rs
```diff
@@ -146,6 +146,86 @@
...
+//! }
+//!
+//! let mut buf = [0u8; 100];
+//! let writer = SliceWrite::new(&mut buf[..]);
+//! let mut ser = Serializer::new(writer);
+//! let user = User {
+//!     user_id: 42,
+//!     password_hash: [1, 2, 3, 4],
+//! };
+//! user.serialize(&mut ser).unwrap();
```
pyfisch: One minor note, can you change the example, so that unwrap isn't used? Thanks.

[Examples use ?, not try!, not unwrap](https://rust-lang-nursery.github.io/api-guidelines/documentation.html#examples-use--not-try-not-unwrap-c-question-mark)

src/lib.rs
```diff
@@ -146,6 +146,86 @@
...
+//! }
+//!
+//! let mut value = [
+//!     0xa2, 0x67, 0x75, 0x73, 0x65, 0x72, 0x5f, 0x69, 0x64, 0x18, 0x2a, 0x6d,
+//!     0x70, 0x61, 0x73, 0x73, 0x77, 0x6f, 0x72, 0x64, 0x5f, 0x68, 0x61, 0x73,
+//!     0x68, 0x84, 0x1, 0x2, 0x3, 0x4
+//! ];
+//! let mut scratch = [0u8; 32];
+//!
+//! use serde_cbor::de::from_slice_with_scratch;
```
pyfisch: A note when to use which one `from_slice_with_scratch` and `from_mut_slice` would be great, but I can also open an issue for this.

src/lib.rs
baloo: the doc example are expected to return Unit, and not a result, which results in this error:
```

---- src/lib.rs -  (line 166) stdout ----
error[E0277]: the `?` operator can only be used in a function that returns `Result` or `Option` (or another type that implements `std::ops::Try`)
  --> src/lib.rs:185:1
   |
22 | user.serialize(&mut ser)?;
   | ^^^^^^^^^^^^^^^^^^^^^^^^^ cannot use the `?` operator in a function that returns `()`
   |
   = help: the trait `std::ops::Try` is not implemented for `()`
   = note: required by `std::ops::Try::from_error`
```

I can put that in a function, but that makes the example less readable in my opinion.

pyfisch: Have a look at the linked doc. You can hide the `fn main() -> Result<...>` in the HTML docs by adding a # to the start of the line.

baloo: Ha, sorry, my mistake.

src/lib.rs
baloo: is this better? not sure how I can explain it any better :/

pyfisch: It's good. :+1:

PR #126, opened 2019-06-14
packed encoding deserialization can now be turned off

### LINKED ISSUE #112: What if I don't want the packed serialization?
I'm trying to avoid having a format that inadvertently works. While avoiding serialization is a matter of just not calling the `_packed` functions, deserialization seems to implicitly always use it.

E.g., coding against a spec, I do not want other, incorrect (according to the spec, which might be agnostic to Rust/`serde_cbor`) to successfully deserialize, lest someone thinks that format *is* supported.

baloo:
Fixes #112 
See #119

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -443,10 +443,12 @@ where
         V: de::Visitor<'de>,
     {
         let accept_packed = self.accept_packed;
+        let accept_named = self.accept_named;
         self.recursion_checked(|de| {
             let value = visitor.visit_map(MapAccess {
                 de,
                 len: &mut len,
+                accept_named,
                 accept_packed,
             })?;
 
@@ -462,9 +464,14 @@ where
     where
         V: de::Visitor<'de>,
     {
+        let accept_named = self.accept_named;
         let accept_packed = self.accept_packed;
         self.recursion_checked(|de| {
-            let value = visitor.visit_map(IndefiniteMapAccess { de, accept_packed })?;
+            let value = visitor.visit_map(IndefiniteMapAccess {
+                de,
+                accept_packed,
+                accept_named,
+            })?;
             match de.next()? {
                 Some(0xff) => Ok(value),
                 Some(_) => Err(de.error(ErrorCode::TrailingData)),
@@ -494,6 +501,7 @@ where
     where
         V: de::Visitor<'de>,
     {
+        let accept_named = self.accept_named;
         let accept_packed = self.accept_packed;
         self.recursion_checked(|de| {
             let mut len = 1;
@@ -502,6 +510,7 @@ where
                     de,
                     len: &mut len,
                     accept_packed,
+                    accept_named,
                 },
             })?;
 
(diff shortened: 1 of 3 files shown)


pyfisch:
looks like it needs a `cargo fmt --all`

baloo:
Ha! I didn't know the tests were not formated by `cargo fmt`! Thanks!

baloo:
I also added tests for the indefinite length map (`0xbf`)

pyfisch:
LGTM.

For symmetry I also want to support the `accept_named` option (to opt out of named fields). Would you also like to implement this here?

src/de.rs
```diff
@@ -947,6 +969,16 @@ where
         }
         *self.len -= 1;
 
+        match self.de.peek()? {
+            Some(_byte @ 0x00..=0x1b) if !self.accept_packed => {
+                return Err(self.de.error(ErrorCode::WrongStructFormat));
+            }
+            Some(_byte @ 0x40..=0x7f) if !self.accept_named => {
```
baloo: I used range 0x40 which is for byte string here. I felt I needed to mention it.

src/de.rs
```diff
@@ -987,6 +1021,12 @@ where
         K: de::DeserializeSeed<'de>,
     {
         match self.de.peek()? {
+            Some(_byte @ 0x00..=0x1b) if !self.accept_packed => {
+                return Err(self.de.error(ErrorCode::WrongStructFormat))
+            }
+            Some(_byte @ 0x40..=0x7f) if !self.accept_named => {
```
baloo: I used range 0x40 which is for byte string here. I felt I needed to mention it.

baloo:
I moved to a struct to feed arguments to `from_slice_stream_options`. I felt I was loosing expressiveness, and this approach felt more expressive to me. Let me know what you think.

pyfisch:
Using a struct `Options` in the tests is fine with me.

src/de.rs
pyfisch: I'm a bit confused, I thought fields would only use UTF-8 names and not binary names?

baloo:
@pyfisch anything else needed here? or can we merge this one?

pyfisch:
> @pyfisch anything else needed here? or can we merge this one?

https://github.com/pyfisch/cbor/pull/126#discussion_r294075703

src/de.rs
baloo: ha, missed that one! well I'm not too sure. That's one of the reasons I raised this question.

baloo:
Just pushed a last commit to try to settle the argument. Serde is happily using byte string as struct field names instead of UTF8 Strings.
In my opinion, this should not be accepted if we were to disable unpacked deserialization.

WDYT?

pyfisch:
>  In my opinion, this should not be accepted if we were to disable unpacked deserialization.

I agree. Only UTF-8 strings should be accepted. (I guess this was added by serde to support formats that don't differentiate between text and binary strings.)

pyfisch:
If you remove binary strings I can merge it.

baloo:
@pyfisch There

pyfisch:
:tada:

PR #135, opened 2019-07-02
fixup ErrorCode::Io variant for feature unsealed_read_write

### LINKED ISSUE #134: Feature unsealed_read_write should work without std
Run `cargo build --no-default-features --features unsealed_read_write`.

It fails with:

```
error[E0599]: no variant or associated item named `Io` found for type `error::ErrorCode` in the current scope
   --> src/error.rs:53:30
    |
53  |             code: ErrorCode::Io,
    |                              ^^ variant or associated item not found in `error::ErrorCode`
...
263 | pub(crate) enum ErrorCode {
    | ------------------------- variant or associated item `Io` not found here

error: aborting due to previous error

For more information about this error, try `rustc --explain E0599`.
error: Could not compile `serde_cbor`.

To learn more, run the command again with --verbose.
```

@baloo now we know why `Error::Io` is needed.

baloo:
Fixes #134
This partially reverts commit a8be10ec8aa7328c0a601cc6bed18149aca6b9c1.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -130,6 +130,8 @@ impl Error {
             ErrorCode::Message => Category::Data,
             #[cfg(feature = "std")]
             ErrorCode::Io(_) => Category::Io,
+            #[cfg(not(feature = "std"))]
+            ErrorCode::Io => Category::Io,
             ErrorCode::ScratchTooSmall => Category::Io,
             ErrorCode::EofWhileParsingValue
             | ErrorCode::EofWhileParsingArray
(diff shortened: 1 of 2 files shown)


src/error.rs
```diff
@@ -267,6 +269,8 @@ pub(crate) enum ErrorCode {
     Message,
     #[cfg(feature = "std")]
     Io(io::Error),
+    #[cfg(all(not(feature = "std"), feature = "unsealed_read_write"))]
```
pyfisch: I'd suggest to use `#[allow(unused)]` here instead.
(The reasoning is that writing these feature bounds correctly will quickly become too complex)

baloo: sounds fair

PR #137, opened 2019-07-08
Correctly set value::Serializer's is_human_readable

### LINKED ISSUE #136: value::Serializer should set is_human_readable to false
Currently `value::Serializer::is_human_readable` returns `true` (by default as it is not overriden), leading to incorrect serialization when compared to using the regular `serde_cbor` serializer which correctly returns `false`.

This causes problems as serializing via `value::Value` is the only way (in 0.10) to achieve canonical serialization.

kostko:
Fixes #136.

I hope this can go into 0.10.1 soon.

### DIFF
--- a/src/value/ser.rs
+++ b/src/value/ser.rs
@@ -240,6 +240,11 @@ impl serde::Serializer for Serializer {
             map: BTreeMap::new(),
         })
     }
+
+    #[inline]
+    fn is_human_readable(&self) -> bool {
+        false
+    }
 }
 
 pub struct SerializeVec {


pyfisch:
thanks

kostko:
Cool, thanks for merging this quickly. Not sure if there are any other bug fixes scheduled for the next patch release, but I would be very grateful if this could go into a 0.10.1 release or similar soonish :-) Thanks again!

pyfisch:
Released as v0.10.1

PR #141, opened 2019-08-16
pub no_std fmt::write

jacobrosenthal:
We have an [interrupted dma coming](https://github.com/nrf-rs/nrf52-hal/pull/102) into embedded hal shortly that uses a statically allocated pool to write out to dma. Serializers like your recent no_std support (thanks!) are an obvious application.

I can get a mut u8 slice and thus use your existing api just fine, either naively with an extra buffer, or even without but for various reasons with the slice api we have to zero the (maybeuninit) buffer which is a little extra overhead thats seems unnecessary . 

It DOES [implement fmt::Write](https://github.com/nrf-rs/nrf52-hal/pull/102/files#diff-4d88bb5a92b057ef97d83602b2d76b30R448) which without core::io::write seems to be the only trait no_std can agree upon?

You do have some kind of fmt::Write implemented.. but its not public and I dont understand it.. it just calls back in upon itself.. 
Frankly I barely understand all the sealed + generics + lifetimes, but I made this new FmtWriter that does what Im trying to accomplish with the end result is I can use fmt::write trait instead of a slice
```
        let mut node = UarteDMAPoolNode::new();
        {
            let writer = FmtWriter::new(&mut node);
            let mut ser = Serializer::new(writer);
            &"aabasdf".serialize(&mut ser).unwrap();
        }
        let b = UarteDMAPool::alloc().unwrap().init(node);
        resources.PRODUCER.enqueue(b).unwrap();
        rtfm::pend(interrupt::UARTE0_UART0);
```

Thoughts. Am I missing something about the existing FmtWrite?
Thanks!

### DIFF
--- a/src/write.rs
+++ b/src/write.rs
@@ -192,7 +192,9 @@ impl<'a, W: fmt::Write> Write for FmtWriter<'a, W> {
     type Error = fmt::Error;
 
     fn write_all(&mut self, buf: &[u8]) -> Result<(), Self::Error> {
-        self.0.write_str(core::str::from_utf8(buf).unwrap())
+        // Is safe as internally an &[u8] is needed, but not allowed by the interface
+        self.0
+            .write_str(unsafe { core::str::from_utf8_unchecked(buf) })
     }
 }
 
--- a/src/write.rs
+++ b/src/write.rs
@@ -168,3 +168,33 @@ impl<'a> Write for SliceWrite<'a> {
 
 #[cfg(not(feature = "unsealed_read_write"))]
 impl<'a> private::Sealed for SliceWrite<'a> {}
+
+/// Implements [`Write`](trait.Write.html) for mutable byte slices (`&mut [u8]`).
+///
+/// Returns an error if the value to serialize is too large to fit in the slice.
+#[cfg(not(feature = "std"))]
+#[derive(Debug)]
+pub struct FmtWriter<'a, W: fmt::Write>(&'a mut W);
+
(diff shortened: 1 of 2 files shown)


src/write.rs
```diff
@@ -168,3 +168,33 @@ impl<'a> Write for SliceWrite<'a> {
...
+        FmtWriter(w)
+    }
+}
+
+#[cfg(not(feature = "std"))]
+impl<'a, W: fmt::Write> Write for FmtWriter<'a, W> {
+    type Error = fmt::Error;
+
+    fn write_all(&mut self, buf: &[u8]) -> Result<(), Self::Error> {
+        self.0.write_str(core::str::from_utf8(buf).unwrap())
```
jacobrosenthal: Have to get rid of this unwrap before a merge anyway, but also from_utf8 pukes on structs or anything else interesting.. so not particularly functional atm

pyfisch:
@wildarch can you please have a look?

I have little knowledge about embedded computing so I can't really review this PR. But I think that if you enable the "unsealed_read_write" feature of this crate you should be able to create a `FmtWriter` or similar in your own crate (without changes to serde_cbor).

jacobrosenthal:
I understand.. If we got no_std to agree on a common core::io::write trait in the future would you upstream it in place of std::io::write?

pyfisch:
> If we got no_std to agree on a common core::io::write trait in the future would you upstream it in place of std::io::write?

Probably.

PR #145, opened 2019-09-01
Support 128-bit integers

nhynes:
This PR adds support for `u128` and `i128` in accordance with [§2.4.2 of the CBOR spec](https://tools.ietf.org/html/rfc7049#section-2.4.2).

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -280,7 +280,7 @@ where
         Ok(BigEndian::read_u64(&buf))
     }
 
-    fn parse_bigint<V>(&mut self, visitor: V, signed: bool) -> Result<V::Value>
+    fn parse_bigint<V>(&mut self, signed: bool, visitor: V) -> Result<V::Value>
     where
         V: de::Visitor<'de>,
     {
@@ -890,7 +890,7 @@ where
         match self.peek()? {
             Some(tag @ 0xc2..=0xc3) => {
                 self.consume();
-                self.parse_bigint(visitor, tag == 0xc3 /* signed */)
+                self.parse_bigint(tag == 0xc3 /* signed */, visitor)
             }
             _ => self.parse_value(visitor),
         }
--- a/src/de.rs
+++ b/src/de.rs
@@ -280,24 +280,31 @@ where
         Ok(BigEndian::read_u64(&buf))
     }
 
-    fn try_parse_u128(&mut self) -> Result<Option<u128>> {
-        // ^ Only *try* parse because the value could be a non-u128 bytestring.
+    fn parse_bigint<V>(&mut self, visitor: V, signed: bool) -> Result<V::Value>
+    where
+        V: de::Visitor<'de>,
+    {
         let desc = match self.peek()? {
             Some(desc) => desc,
-            None => return Ok(None),
+            None => return Err(self.error(ErrorCode::EofWhileParsingValue)),
         };
         let ty = desc >> 5;
         if ty != 2 {
-            return Ok(None); // not a bytestring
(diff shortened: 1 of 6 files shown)


src/value/mod.rs
```diff
@@ -25,13 +25,10 @@ pub enum Value {
...
-    /// Integer CBOR numbers.
-    ///
-    /// The biggest value that can be represented is 2^64 - 1.
-    /// While the smallest value is -2^64.
-    /// Values outside this range can't be serialized
-    /// and will cause an error.
+    /// Signed integer CBOR numbers.
     Integer(i128),
+    /// Unsigned integer CBOR numbers.
+    Unsigned(u128),
```
pyfisch: I don't like to have two overlapping integer variants (again) because a postive number can now be stored in two different ways.
Ist there something special about 128-bits that you need `u128` numbers that do not fit into `i128`?

pyfisch (commented):
I am not sure if I want to support big integers. They would be the first supported tagged value, additionally it breaks code that relies on the assumption that big integers are deserialized as bytestrings. (I don't know if there is such code)

src/value/mod.rs
nhynes: > Is there something special about 128-bits that you need u128 numbers that do not fit into i128?

The only issue is with the [canonical sort order](https://github.com/nhynes/cbor/blob/support-i128/src/value/mod.rs#L141). Before, `i128` could record the sign and magnitude of the largest supported number. Since `i128` is the largest data type, the sign isn't recorded independently of magnitude.

nhynes:
> They would be the first supported tagged value

Well, it's in accordance with the spec and doesn't much affect the codebase.

> Additionally it breaks code that relies on the assumption that big integers are deserialized as bytestrings.

Yes, that's a fair point. How about, instead, special-casing the implementation of `Deserialize::deserialize_[iu]128`? Is that a backwards-compatible change that you'd be willing to accept?

pyfisch:
@sfackler What do you think about supporting 128-bit integers in this crate?

src/de.rs
```diff
@@ -280,6 +280,33 @@ where
         Ok(BigEndian::read_u64(&buf))
     }
 
+    fn parse_bigint<V>(&mut self, signed: bool, visitor: V) -> Result<V::Value>
```
pickfire: Is it better to separate this as a function rather than just keeping it down?

grantperry:
@nhynes how hard would it be to fix your branch to include #206? It seems you may be able to wrap all ui128 functions in `serde_if_integer! {...}`?

surban:
Is there something that we can do to move this forward? Being unable to serialize a basic Rust type unfortunately makes serde_cbor not usable for me.

pyfisch:
> Is there something that we can do to move this forward?

You can pay me to work on this crate. Or you can find a company or team that wants to take over the maintenance of this crate.

> Being unable to serialize a basic Rust type unfortunately makes serde_cbor not usable for me.

128-bit integers aren't part of CBOR (big ints are however) and 128-bit integers weren't part of Rust 1.0 as well. You can always write a custom (de)serializer for your data structures.

There is another serde CBOR crate: https://github.com/enarx/ciborium maybe it is right for your use-case. (They handle 128-bit integers.)

pickfire:
> > Is there something that we can do to move this forward?
>
> You can pay me to work on this crate. Or you can find a company or team that wants to take over the maintenance of this crate.

Or you can wait until someone who is interested to pick this up. At least I am not interested in this right now.

PR #147, opened 2019-09-02
Implement 'to_vec' and 'from_slice' for no_std

geigerzaehler:
We implement `serde_cbor::to_vec()` and `serde_cbor::from_slice()` in a `no_std` environment to simplify serialization. For this we use `alloc::vec::Vec`.

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -15,11 +15,12 @@ use crate::error::{Error, ErrorCode, Result};
 use crate::read::EitherLifetime;
 #[cfg(feature = "unsealed_read_write")]
 pub use crate::read::EitherLifetime;
-use crate::read::Offset;
 #[cfg(feature = "std")]
-pub use crate::read::{IoRead, SliceRead};
+pub use crate::read::IoRead;
+use crate::read::Offset;
+#[cfg(any(feature = "std", feature = "alloc"))]
+pub use crate::read::SliceRead;
 pub use crate::read::{MutSliceRead, Read, SliceReadFixed};
-
 /// Decodes a value from CBOR data in a slice.
 ///
 /// # Examples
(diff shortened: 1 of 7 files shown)


geigerzaehler:
Unfortunately this does not build with Rust 1.31.0 because the `alloc` crate is not available. I’d suggest not running the `--no-default-features` tests on Rust 1.31.0 and documenting that this only works with newer versions of Rust. Alternatively we could introduce an `alloc` flag (like serde) so that `no_std` still works with Rust `1.31.0`.

pyfisch:
Is it correct that with this change alloc is always used?
I *think* there are people using CBOR on systems which do not support allocation.

geigerzaehler:
> Is it correct that with this change alloc is always used?

I think so, though I’m not super familiar with `alloc` to be honest. 

> I _think_ there are people using CBOR on systems which do not support allocation.

Ok. It might be best then to put this behind a feature flag. This also aligns with serde and retains compatibility with 1.31.0. I’ll update the PR accordingly.

pyfisch:
cc @wildarch @chrysn

src/de.rs
```diff
@@ -15,13 +15,16 @@ use crate::error::{Error, ErrorCode, Result};
...
-pub use crate::read::{IoRead, SliceRead};
+pub use crate::read::IoRead;
+use crate::read::Offset;
+#[cfg(any(feature = "std", feature = "alloc"))]
+pub use crate::read::SliceRead;
 pub use crate::read::{MutSliceRead, Read, SliceReadFixed};
-
 /// Decodes a value from CBOR data in a slice.
 ///
+/// Only available with the `std` (default) or `alloc` feature.
```
wildarch: Bit of a nit: I think this could be omitted, as presumably 99% of users will just use the `std` version. The remaining 1% that do care about `no_std` can build the documentation without the `std` feature to see that this function is unavailable (or browse the source).

geigerzaehler: Makes sense. I remove these comments.

pyfisch:
@geigerzaehler  By the way this PR broke Travis CI. :wink: See: https://github.com/pyfisch/cbor/pull/154 and https://stackoverflow.com/questions/58249675/travis-ci-fails-to-parse-travis-yml

geigerzaehler:
Sorry and thanks for fixing. I was wondering where the green checkmark was. To bad Travis does not fail the build when it can’t parse `.travis.yaml`.

PR #148, opened 2019-09-03
Don't abort build sucessfully when clippy fails

geigerzaehler:
Before, if the `clippy` check would fail `exit 0` would be called which exited the test script. This meant that all subsequent steps would be skipped and the _build would pass_.

Now instead of exiting the test script we just make sure that the script always return a zero exit code.


pyfisch:
Thanks for catching this!

PR #151, opened 2019-09-25
Add tags support behind feature flag

vmx:
Tags are stored as Newtype Structs with the special name `_TagStruct`. The tags
are represented as a struct tuple consisting of the tag identifier and a value.

Example:

```rust
let tag = Value::Tag(123, Box::new(Value::Text("some value".to_string())));
let tag_encoded = to_vec(&tag).unwrap();
```

This implementation is heavily based on the msgpack-rust library, when support
for extensions were added [1][2]. They have the same problem as CBOR has, that
such an extension/tagging type is not directly supported by Serde, hence a
workaround is needed. For me it makes sense to use a similarly working workaround
across different libraries.

This commit also contains an [example](../blob/05eabb4526ecb257e60f85b3ae4739e80ea1e6b5/examples/tags.rs) on how to use custom types as tags. It can
be run via:

    cargo run --example tags --features tags

This is an alternate proposal to https://github.com/pyfisch/cbor/pull/129.

[1]: https://github.com/3Hren/msgpack-rust/commit/a34ab8fcca9f1133ec1c4fc446a59b71acf5613e
[2]: https://github.com/3Hren/msgpack-rust/pull/216

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,3 +21,4 @@ script:
   - [[ $TRAVIS_RUST_VERSION != "1.31.0" ]] && cargo build --no-default-features --features alloc
   - cargo build --features unsealed_read_write # The crate should still build when the unsealed_read_write feature is enabled.
   - cargo build --no-default-features --features unsealed_read_write # The crate should still build when the unsealed_read_write feature is enabled and std disabled.
+  - cargo test --features tags # Run tags tests
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ serde = { version = "1.0.14", default-features = false }
 
 [dev-dependencies]
 serde_derive = { version = "1.0.14", default-features = false }
+serde_bytes = "0.11.2"
 
 [features]
 default = ["std"]
@@ -31,3 +32,4 @@ default = ["std"]
 alloc = ["serde/alloc"]
 std = ["serde/std" ]
 unsealed_read_write = []
+tags = []
--- a/examples/tags.rs
+++ b/examples/tags.rs
@@ -1,0 +1,125 @@
+fn main() {
+    #[cfg(feature = "tags")]
+    tags_example::main();
+    #[cfg(not(feature = "tags"))]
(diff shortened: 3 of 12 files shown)


vmx:
I really thought this is a good idea. But after digging into all the discussions about tag support in Serde, I also come to the conclusion that this should be properly solved within Serde so that there's interoperability between different formats.

I would want to revive the approach take at https://github.com/serde-rs/serde/pull/408. @pyfisch do you still have your `tagged_values` branch of cbor lying around somewhere (no worries if not, it shouldn't be too hard to implement)?

pyfisch:
Hi @vmx, integrating with serde is a great idea. The `tagged_values` branch is still available on GitHub isn't it? https://github.com/pyfisch/serde/tree/tagged_values

I'd probably do some things differently than in this old branch. For example I'd use a `TaggedVisitor` as suggested in [the discussion](https://github.com/serde-rs/serde/pull/408/files#r68238565).

pyfisch:
I will send a PR for the serialization part to the serde repo soon.

vmx:
@pyfisch Wow, that's great to hear. I so far have ported the old version to master and I started playing around with it. I thought about creating a small example with a tag that is used in both, CBOR and msgpack, which is then (correctly) serialized/deserialized.

So please let me know if the new version of the patch for Serde is up. I really want to see this happen and I also have time to spend on this if there's a chance to get this upstream :)

pyfisch:
See also #157

vmx:
@pyfisch I haven't done that much, I pushed whatever I currently have though:

 - Serde: https://github.com/vmx/serde/tree/wip-revive-tagged-values
 - CBOR: https://github.com/vmx/cbor/tree/wip-revive-tagged-values

I kind of stopped when I saw that you're working on it again. But I'm happy to help, when you tell me what I should work on (my Serde knowledge is still pretty basic, though).

rklaehn:
Tags are on master, so I guess we can close this.

PR #156, opened 2019-10-05
Undo .travis.yml changes

pyfisch:
See #155


pyfisch:
the added line doesn't work

PR #159, opened 2019-10-18
Enable byte_offset for IoRead and SliceReadFixed

### LINKED ISSUE #158: byte_offset not available for file reader
Thank you for your work on serde_cbor, it is quite useful.

# Description
I currently have a log-structured file containing independently written cbor encoded objects.
I'm trying to iterate through the file, reading and decoding all the objects, and noting the offset of each object in the file. This is supposed to allow later lookup without keeping the whole object in memory. Unfortunately I can get the `byte_offset` function to work. Apparently the Offset trait isn't implemented for a reader using a file.
Is there an easier way to read all file contents and get the offsets of each object? Am I doing something wrong with with the the deserializer?

# Compiler Error Message
```
error[E0599]: no method named `byte_offset` found for type `serde_cbor::de::Deserializer<serde_cbor::read::IoRead<std::fs::File>>` in the current scope
   --> src/lib.rs:152:26
    |
152 |             offset = des.byte_offset() as u64;
    |                          ^^^^^^^^^^^
    |
    = note: the method `byte_offset` exists but the following trait bounds were not satisfied:
            `serde_cbor::read::IoRead<std::fs::File> : serde_cbor::read::Offset`

```

# Example code
```rust
let mut des = serde_cbor::Deserializer::from_reader(file);
let offset = des.byte_offset() as u64;
```

# Context
I'm working through a [rust course](https://github.com/pingcap/talent-plan/blob/master/rust/projects/project-2/project.md). One of the steps is keeping the offsets for each key-value-pair in the file in memory to access the value later.

pyfisch:
Implement the Offset trait for both types.
Do not implement Offset for MutSliceRead as the
offset may change.

closes #158

@baloo Would you take a look?

### DIFF
--- a/src/read.rs
+++ b/src/read.rs
@@ -264,6 +264,16 @@ where
 }
 
 #[cfg(feature = "std")]
+impl<R> Offset for IoRead<R>
+where
+    R: std::io::Read,
+{
+    fn byte_offset(&self) -> usize {
+        self.offset() as usize
+    }
+}
+
+#[cfg(feature = "std")]
 #[derive(Debug)]
 struct OffsetReader<R> {
     reader: R,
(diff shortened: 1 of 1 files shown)


baloo:
@pyfisch ha, sorry I missed the notification. That did look good to me.

PR #162, opened 2019-10-30
Use `slice::iter` instead of `into_iter` to avoid future breakage

LukasKalbertodt:
`an_array.into_iter()` currently just works because of the autoref
feature, which then calls `<[T] as IntoIterator>::into_iter`. But
in the future, arrays will implement `IntoIterator`, too. In order
to avoid problems in the future, the call is replaced by `iter()`
which is shorter and more explicit.

A crater run showed that your crate is affected by a potential future 
change. See https://github.com/rust-lang/rust/pull/65819 for more information.

### DIFF
--- a/tests/canonical.rs
+++ b/tests/canonical.rs
@@ -22,7 +22,7 @@ mod std_tests {
             -65537,
             -4294967296,
         ]
-        .into_iter()
+        .iter()
         .map(|i| Value::Integer(*i))
         .collect::<Vec<_>>();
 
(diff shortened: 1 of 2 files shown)


pyfisch:
Thank you!

LukasKalbertodt:
@pyfisch Not to rush anyone, but could this change be released as 0.10.3? Lots of crates depend on this one, so getting this fix out soon would be great :)

pyfisch:
Done.

But I doubt that this will help much as many crates still depend on v0.9 and some even on v0.8 and earlier releases.

LukasKalbertodt:
@pyfisch Thanks a bunch!

Unfortunately, I was not particularly smart regarding this PR. The same happened for `itertools`. This PR only fixes tests and when users depend on this crate, these tests are not compiled. So indeed not a single user crate regressed because of this crate. This is good in a sense, but also bad since I caused you unnecessary work and a useless release. I am truly sorry for that! 

At least we are completely done here now.

PR #164, opened 2019-11-16
Implement tagged values by using a thread-local

rklaehn:
The problem with serde is that there is a communication channel missing between serializers/deserializers and visitors.

The basic idea of this PR is to use a tiny thread local value as this communications channel (until we hopefully get some more help from serde in the future).

# Deserialization

When we find a tag, we store it in a thread-local of type `Option<u64>` and then call visit_newtype_struct. In a visitor that understands tags, *only in visit_newtype_struct*, we get the tag and wrap the resulting value in a `Value::Tag` if appropriate. Essentially this is exactly the same as if there was an additional u64 argument cbor_tag.

# Serialization

For serialization the approach is the same. A `Value::Tag` sets the tag, and the serializer consumes the tag and emits the tag on the cbor binary stream, then continues.

# Custom serializers

For custom serializers, there is just an extension method on serde::ser::Serializer that allows to serialize with a cbor tag prefix:

```rust
struct Cid(Vec<u8>);
impl Serialize for Cid {
    fn serialize<S>(&self, s: S) -> Result<S::Ok, S::Error>
    where
        S: Serializer,
    {
        s.serialize_cbor_tagged(42, &self.0)
    }
}
```

For deserialisation, there is an optional method to check the right tag:

```rust
impl<'de> Deserialize<'de> for Cid {
    fn deserialize<D>(deserializer: D) -> Result<Cid, D::Error>
    where
        D: Deserializer<'de>,
    {
        deserializer.expect_cbor_tag(42)?;
        let res = Vec::<u8>::deserialize(deserializer)?;
        Ok(Cid(res))
    }
}
```

But using this has the downside that deserializing CBOR that has been written with a non tag enabled version of serde_cbor will not work.

# Problems

There are tons of not very nice things about this approach compared to having a parameter to pass through the info. E.g. the tag will remain set until somebody consumes it. Also, you would need separate tags for serialize and deserialize for something like serde_transcode to work.

Nevertheless the simplicity of this hack is appealing...

# No-Std

Thread locals are part of the std lib and not in core, so this hack won't work with no-std. But it would be behind a feature flag anyway, so that might not be a big deal.

# To  do

- [ ] integrate tags feature flag with the rest
- [ ] ensure tests run with `tags` feature
- [ ] documentation


rklaehn:
@pyfisch added a test as requested in https://github.com/rklaehn/cbor/pull/2#issuecomment-554747762

seems to work.

pyfisch:
Tests are failing

rklaehn:
No idea what is wrong there. It is just the 1.31.0 tests that fail. I downgraded to 1.31.0 and cargo test passes.

rklaehn:
@vmx WDYT? Really not nice, but it works, and I don't see nicer options in the current situation...

vmx:
Looks like a good solution to me. I don't really have any preference of one solution/hack to another.

rklaehn:
The good thing about this approach is that the thread local is outside normal data values, so a deserializer that does not know them will just ignore them completely.

rklaehn:
So, questions before this can be merged:

Should this be hidden behind a feature flag? I would say yes.
Should the serde_cbor::Value::Tag be hidden behind a feature flag? Currently it is not behind a feature flat, but serializing a tag will be a noop. Maybe also put it behind the feature flag?
What about the extension methods for serde serializer and deserializer? Good idea or not?

src/lib.rs
```diff
@@ -308,3 +309,8 @@ pub use crate::ser::to_writer;
 #[cfg(feature = "std")]
 #[doc(inline)]
 pub use crate::value::Value;
+
```
pyfisch: Why do you need these use statements? lib.rs should only include the public re-exports.

examples/dagcbor.rs
```diff
@@ -0,0 +1,53 @@
+use serde::de::{Deserialize, Deserializer};
+use serde::ser::{Serialize, Serializer};
+use serde_cbor::{DeserializerExt, SerializerExt, Value};
+use std::error::Error;
+use std::fs::File;
+use std::io::Cursor;
+
+#[derive(Debug, PartialEq)]
+struct Cid(Vec<u8>);
```
pyfisch: What are "dagcbor" a "Cid"?

examples/dagcbor.rs
```diff
@@ -0,0 +1,53 @@
...
+
+fn main() -> Result<(), Box<dyn Error>> {
+    let mut file = File::open("examples/test.cbor")?;
+    let mut bytes = Vec::new();
+    std::io::copy(&mut file, &mut bytes)?;
+    println!("{:?}", bytes);
+    let ast: serde_cbor::Value = serde_cbor::from_reader(Cursor::new(bytes))?;
+    println!("{:?}", ast);
+    let bytes1 = serde_cbor::to_vec(&Value::Tag(42, Box::new(Value::Bytes(Vec::new()))))?;
+    // let bytes = serde_cbor::to_vec(&Value::Null)?;
```
pyfisch: Please remove.

src/de.rs
```diff
@@ -397,6 +398,21 @@ where
         }
     }
 
+    fn recursion_checked_tagged<F, T>(&mut self, tag: u64, f: F) -> Result<T>
```
pyfisch: This function is only called like this: `self.recursion_checked_tagged(tag.into(), |de| visitor.visit_newtype_struct(de))`. Therefore we don't need to pass in f but can just include it in the function.
Furthermore it should call `recursion_checked` and not reimplement it.

The new function should look like this: `fn handle_tagged_value<F, T, V>(&mut self, tag: u64, visitor: V) -> Result<V::Value>`

src/de.rs
```diff
@@ -704,22 +720,25 @@ where
             0xbf => self.parse_indefinite_map(visitor),
 
             // Major type 6: optional semantic tagging of other major types
-            0xc0..=0xd7 => self.recursion_checked(|de| de.parse_value(visitor)),
+            0xc0..=0xd7 => {
+                let tag = byte as usize - 0xc0;
```
pyfisch: Instead: `let tag = byte as u64 - 0xc0;` (Avoid converting from u8 to usize to u64)

src/de.rs
```diff
@@ -704,22 +720,25 @@ where
...
-                self.parse_u32()?;
-                self.recursion_checked(|de| de.parse_value(visitor))
+                let tag = self.parse_u32()?;
+                self.recursion_checked_tagged(tag.into(), |de| visitor.visit_newtype_struct(de))
             }
             0xdb => {
-                self.parse_u64()?;
-                self.recursion_checked(|de| de.parse_value(visitor))
+                let tag = self.parse_u64()?;
+                self.recursion_checked_tagged(tag as u64, |de| visitor.visit_newtype_struct(de))
```
pyfisch: `tag as u64` is unnessary?

src/tagstore.rs
```diff
@@ -0,0 +1,111 @@
+use serde::de::{Deserializer, Error};
+use serde::ser::{Serialize, Serializer};
+
+/// signals that a newtype is from a CBOR tag
+pub const CBOR_NEWTYPE_NAME: &str = "__cbor_tag";
```
pyfisch: This should include some special char to make sure this is *never* the actual name of the struct. For example include `\0` or `\x1b` at the start.

src/tagstore.rs
```diff
@@ -0,0 +1,111 @@
...
+            Some(_) => Err(Self::Error::custom("unexpected cbor tag")),
+            None => Err(Self::Error::custom("missing cbor tag!")),
+        }
+    }
+}
+
+impl<'de, D: Deserializer<'de>> DeserializerExt<'de> for D {}
+
+#[cfg(tags)]
+pub mod tag_access {
```
pyfisch: So code above is always included but this is only included if the `tags` feature is enabled, right?
Shouldn't the code above also be disabled if tags are not used?

src/tagstore.rs
```diff
@@ -0,0 +1,111 @@
...
+            *f.borrow_mut() = value;
+        });
+    }
+
+    pub fn get_tag() -> Option<u64> {
+        CBOR_TAG.with(|f| *f.borrow())
+    }
+
+    #[cfg(test)]
+    mod tests {
```
pyfisch: Currently all tests are in the tests/ folder. I want to keep it that way.

src/tagstore.rs
```diff
@@ -0,0 +1,111 @@
...
+
+/// extensions for all serde serializers
+pub trait SerializerExt: Serializer {
+    /// basically serialize_newtype_struct with a cbor tag value
+    fn serialize_cbor_tagged<T: Serialize>(
+        self,
+        tag: u64,
+        value: &T,
+    ) -> core::result::Result<Self::Ok, Self::Error> {
+        tag_access::set_tag(Some(tag));
```
pyfisch: One idea instead of having `set_tag` and `get_tag` would be that the serializer consumes the tag previously stored. This would avoid the need for an explicit `set_tag(None)` and assures that each tag is only used once. Is this possible?

src/value/de.rs
```diff
@@ -134,6 +135,18 @@ impl<'de> de::Deserialize<'de> for Value {
             {
                 Ok(Value::Float(v))
             }
+
+            fn visit_newtype_struct<D>(self, deserializer: D) -> Result<Self::Value, D::Error>
+            where
+                D: serde::Deserializer<'de>,
+            {
+                let tag = get_tag();
+                let result = deserializer.deserialize_any(self);
```
pyfisch: Can't you write `let value = deserializer.deserialize_any(self)?;` instead?

tests/de.rs
```diff
@@ -220,10 +220,24 @@ mod std_tests {
         assert_eq!(value.unwrap(), Value::Float(100000.0));
     }
 
+    #[cfg(tags)]
     #[test]
     fn test_self_describing() {
         let value: error::Result<Value> =
```
pyfisch: Lets change this instead to `Result<String>`, so only a string is expected an we don't need a different test depending on whether tags are enabled.

pyfisch:
> Should this be hidden behind a feature flag? I would say yes.

I'm not sure. Do we expect any problems if it is enabled by default?
(One way to go about this is to add `tags` as a default enabled feature in v0.11 and change it to opt-in if problems arise in the next release)

> Should the serde_cbor::Value::Tag be hidden behind a feature flag? Currently it is not behind a feature flat, but serializing a tag will be a noop. Maybe also put it behind the feature flag?

It should be always there. But maybe add a note to the documentation.

> What about the extension methods for serde serializer and deserializer? Good idea or not?

Can you remove it from this PR? We can maybe include it in a future PR.

examples/dagcbor.rs
rklaehn: This is just a test. I need tags support for IPFS/IPLD, just like @vmx and @dignifiedquire... https://docs.ipfs.io/guides/concepts/cid/

Can change this to another example if you want.

src/tagstore.rs
rklaehn: I just wanted to get it to build somehow first. Will change.

src/tagstore.rs
rklaehn: OK, will move this.

src/tagstore.rs
rklaehn: Yes. The reason I unset the tag is that I don't know that the serializer on "the other side of the fence" is a CBOR aware serializer. If it is not, the serializer would *not* consume the tag and it would remain set on this thread.

Then later on the same thread you could have a CBOR deserializer being called from a non CBOR format, and it would read the tag value.

This is far fetched, but possible. So setting the tag back to None seems safer.

rklaehn:
Thanks for the feedback. Not sure when I will get to working on this, since I am traveling over the weekend.

> > Should this be hidden behind a feature flag? I would say yes.
>
> I'm not sure. Do we expect any problems if it is enabled by default?
> (One way to go about this is to add tags as a default enabled feature in v0.11 and change it to opt-in > if problems arise in the next release)

I am not sure about the implications of thread locals. Tried to find out more, but so far did not get very precise answers. They seem to be pretty lightweight though. https://godbolt.org/z/MZDnX4

The whole thread local approach feels a bit dirty compared to having a direct way to pass the info. But we have to work with what serde gives us, which is not much... Also, the user should not notice any of this. And tags *are* an integral part of CBOR...

> > Should the serde_cbor::Value::Tag be hidden behind a feature flag? Currently it is not behind a feature flat, but serializing a tag will be a noop. Maybe also put it behind the feature flag?
>
> It should be always there. But maybe add a note to the documentation.

That's what I thought. Tags are part of CBOR spec. On the other hand somebody might work with tags with the feature accidentally disabled and then wonder why they are not properly serialized...

> > What about the extension methods for serde serializer and deserializer? Good idea or not?
>
> Can you remove it from this PR? We can maybe include it in a future PR.

You need at least *some* way to serialize a tag from a custom serializer for a struct, otherwise the whole PR is not usable for what I intend to use it for. The question is whether it should be an extension method or just a function.

The method that checks if the tag is the correct one is more debatable, but I think it is still useful. E.g. if you want to write deserializers for some of the canonical CBOR tags. Maybe just plain access to the tag would be best...

https://www.iana.org/assignments/cbor-tags/cbor-tags.xhtml

src/tagstore.rs
rklaehn: (note that if the [visit_newtype_struct](https://docs.serde.rs/serde/de/trait.Visitor.html#method.visit_newtype_struct) would take a name parameter, I would not have to set the tag to None and in fact it would not even have to be an option. I could just check if the name parameter in both serialisation and deserialisation and be 100% sure that the caller is serde_cbor and has just set the tag...

But unfortunately it does not...

Some kind of extension mechanism in serde itself sure would be nice, but it seems we won't get one any time soon...

tests/de.rs
rklaehn: I don't get it. If I would stringify the value using Debug, you would still see the Tag(...) when tags are  enabled and not see the tag when they are disabled. You mean json stringify?

src/tagstore.rs
rklaehn: I wonder if serde would accept something like

```
Visitor:
visit_extra_data(kind: u64, data: &[u8])

Serializer:
serialize_extra_data(kind: u64, data: &[u8])
```

That would be very low level, but generic enough for a lot of use cases (json whitespace, tags of all sorts, ...). There would have to be some repo where people can reserve a kind id, but that would be outside of serde...

rklaehn:
A good test for the usability of this would be to write something like serde_cbor::Value that supports one or two of the canonical extensions from https://www.iana.org/assignments/cbor-tags/cbor-tags.xhtml

if you can write such a thing without having to touch the guts of the library, it is sufficiently expressive.

Do you have examples of cbor containing such extensions? The only extension I care about is the one from IPFS, but other extensions might be more relevant to users of this library...

rklaehn:
So, I made a few of the suggested changes. I moved all tags related stuff into a submodule `tags`, made the suggested changes to the deserialiser (much better now) and used the SerializerExt and DeserializerExt methods to implement the tag part of serde_cbor::Value.

I also no longer export SerializerExt and DeserializerExt at top level, but just in tags.

Does this look like it is going in the right direction, @pyfisch ? If so I would polish it a bit, add more docs and examples, and provide better explanations for the dagcbor.rs example. Then as far as I am concerned it would be good to go.

examples/dagcbor.rs
pyfisch: > Can change this to another example if you want.

Maybe this is easier than explaining the example for people who don't know IPFS. For example show how to decode this message from a "bookmark service". It has tagged values for dates and URLs. 

```
{
"title": "The Example Domain",
"link": 32("http://example.org/"),
"created": 0("2003-12-13T18:30:02Z")
}
```

tests/de.rs
pyfisch: I don't want this test for the self-describe tag to vary depending on whether tags are enabled.

```rust
let value: error::Result<String> = de::from_slice(&[0xd9, 0xd9, 0xf7, 0x66, 0x66, 0x6f, 0x6f, 0x62, 0x61, 0x72]);
assert_eq!(value.unwrap(), "foobar".to_owned());
```

pyfisch:
I'm sorry that it takes me so long to reply.

rklaehn:
No problem about the reply time. I know that other things sometimes take priority. Will look into this later!

rklaehn:
FYI I was working on wrapping this up yesterday and hit a snag with reading of tagged values. I think I have a nicer solution to the SerializerExt and DeserializerExt thingies now, but it will take me some time to polish this up. Maybe on the weekend.

rklaehn:
OK, so here is the new approach in a nutshell:

- public access to the thread local tag is removed, as are the `SerializerExt` and `DeserializerExt`. Instead all serialization and deserialization goes via a helper `struct Tagged`. This struct is the complete public interface of the tags feature.

- If you want a newtype `T` that adds a tag, you have to explicitly serialize/deserialize via Tagged<T>. I think this is not that painful to use. Look at the new example to see how this works.

- The Tagged struct is also used in the serde_cbor::Value serializer and deserializer.

One thing I have not yet managed to do is to make it possible for a newtype writer to accept the case where the tag is not present. For that, you would have to somehow have a visitor that intercepts the newtype, but passes though all the other things. I have found that very boilerplatey. Maybe you have an idea.

But I think the whole thing is already useful enough as it is to be released. You can deal with "optional" tags using serde_cbor::Value and with some effort also in your own newtypes.

rklaehn:
OK, after even more serde boilerplate I got the ```tags::Tagged``` roundtrip to json to work.

If you read a `Tagged<T>` from a format that does not have tags, it will still work, but the `tag` value of the `Tagged<T>` will be `None`. Then you can decide what to do - fail with a "tag missing" error or allow it.

In the example I allow it, so you can have a json roundrip....

rklaehn:
So now everything works that I wanted to work, so for me it is fine now. Should I open a new PR so we don't always have to scroll though all this discussion?

Note that this is still a pretty big change. E.g. a tagged string would before deserialize as a string, because the tags are "invisible". Now there is this call to `visit_newtype_struct` for the tag, so a tagged string *will no longer deserialize* as a string. There is no way to avoid this with reasonable effort. Just so you are aware @pyfisch .

That might influence the decision whether to enable tags by default...

vmx:
@rklaehn Thanks for all this hard work. I've tried it out for the use-case for tags that I have and it looks good so far: https://github.com/vmx/rust-ipld-dag-cbor.

tests/de.rs
pyfisch: @rklaehn This is not what I wanted. Doesn't the code work I provided in the previous comment?

rklaehn: Unfortunately not. If you compile with tags enabled, a tagged string no longer serialises as a string. The tag causes the `visit_newtype_struct` to be called, which will throw an error for the string deserializer.

See my comment at the end of the discussion thread.

rklaehn:
So what do we do with this now? I got a few days left to work on this before work starts again. Would love to get this merged until then, as a default or optional feature.

pyfisch:
>  making tags completely "invisible" for types that are unaware of tags is not possible.

This is kind of a deal-breaker.

> Unfortunately not. If you compile with tags enabled, a tagged string no longer deserialises as a string. The tag causes the visit_newtype_struct to be called, which will throw an error for the string deserializer.

 I was under the impression that previously tags did not interfere with normal deserialization. Has this PR always had this issue?

pyfisch:
If tags aren't ignored during normal deserialization one needs to check *everywhere* for tags. Unless of course tags are forbidden everywhere in a protocol except in a few known locations. I think this is rather impractical. Even if tags are behind a feature-gate a single crate in the dependency tree using the feature would break all other crates not expecting tags, right?

rklaehn:
> > making tags completely "invisible" for types that are unaware of tags is not possible.
> 
> This is kind of a deal-breaker.
> 


> > Unfortunately not. If you compile with tags enabled, a tagged string no longer deserialises as a string. The tag causes the visit_newtype_struct to be called, which will throw an error for the string deserializer.
> 
> I was under the impression that previously tags did not interfere with normal deserialization. Has this PR always had this issue?

Yes, a tag is a kind of newtype. So when you deserialize a tag it calls visit_newtype_struct. If you are in control of the deserializer you can work around it. But if not there is not much you can do as far as I can see.

> If tags aren't ignored during normal deserialization one needs to check _everywhere_ for tags. Unless of course tags are forbidden everywhere in a protocol except in a few known locations. I think this is rather impractical. Even if tags are behind a feature-gate a single crate in the dependency tree using the feature would break all other crates not expecting tags, right?

I don't really know what happens if you have multiple dependencies with different feature sets. I guess they will be

pyfisch:
> By the way: I understand if you don't want to merge this. Maybe it is best to have this as a fork.

Many people clearly need tags therefore I now think it should be included in the crate. Albeit behind a feture gate. Would you mind submitting a new PR with the final changes needed, then I will review it, merge it and release v0.11 with optional tag support.

rklaehn:
OK, will do so tomorrow. Sounds like a good compromise.

PR #165, opened 2019-11-24
Use three integer Value variants

### LINKED ISSUE #140: Value i128 serializer breaks cbor <=> json Value to Value conversions
Hello.

Starting with 8cfcea9326e2e6bdf645e897c09af1c533f584c5, direct conversions between cbor `Value`s and json `Value`s is no longer possible.

I fully understand the rationale behind using one integer type to cover the full integer value range supported by cbor. But I wonder what are your thoughts on bringing back, what was admittedly accidental, support for this type of conversion.

MoSal:
* `UnsignedInteger(u64)`: for all non-negative values.
 * `SignedInteger(i64)`: for negative values that fit in a i64.
 * `LargeSignedInteger(i128)`: for negative values that can't be represented by i64.

 This should fix #140.

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -1209,7 +1209,7 @@ where
 /// ];
 /// let mut it = Deserializer::from_slice(&data[..]).into_iter::<Value>();
 /// assert_eq!(
-///     Value::Integer(1),
+///     Value::UnsignedInteger(1),
 ///     it.next().unwrap().unwrap()
 /// );
 /// assert_eq!(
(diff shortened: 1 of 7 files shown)


pyfisch:
Sorry this is not what I meant. I should have communicated this more clearly.

What I meant is to change [the serialization for integer values](https://github.com/pyfisch/cbor/blob/master/src/value/ser.rs#L23) to serialize to an `u64` or `i64` instead of an `i128` if possible.

pyfisch:
I will make this change soon.

PR #168, opened 2019-12-30
WIP: Configurable packing

nickray:
In the FIDO2 CTAP protocol, there are some peculiarities (they have [their own "canonical" format](https://fidoalliance.org/specs/fido-v2.0-ps-20190130/fido-client-to-authenticator-protocol-v2.0-ps-20190130.html#ctap2-canonical-cbor-encoding-form) which slightly differs from RFC 7409).

One requirement is partial "packing", e.g. "AuthenticatorInfo" is a map with integer keys (packed), with values that are maps with unicode string keys. I think packing a struct itself is not expressible via `serde` attributes, which is why it's implemented here?

Another is that their packing starts counting at 1 instead of 0.

<sup>
For example, the message: `Info(versions: ['FIDO_2_0'], aaguid: h'41414755494430313233343536373839', options: {'rk': False, 'up': True, 'plat': False}, max_message_size: 7609)` is expected to serialize as `a40181684649444f5f325f3003504141475549443031323334353637383904a362726bf4627570f564706c6174f405191db9` which is `{1: ['FIDO_2_0'],  3: b'AAGUID0123456789', 4: {'rk': False, 'up': True, 'plat': False}, 5: 7609}`
</sup><br><br>

The intent of this patch is to make these requirements possible.
Thoughts?

I'm very open to better ways of doing this, in particular I'd prefer being able to configure everything directly around the struct definition, which currently is: https://gist.github.com/nickray/284deae46aa1432a0ff89300f97dba94.

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -132,6 +132,8 @@ pub struct Deserializer<R> {
     accept_packed: bool,
     accept_standard_enums: bool,
     accept_legacy_enums: bool,
+    packing_offset: u8,
+    deserializing_struct_key: bool,
 }
 
 #[cfg(feature = "std")]
@@ -190,6 +192,8 @@ where
             accept_packed: true,
             accept_standard_enums: true,
             accept_legacy_enums: true,
+            packing_offset: 0,
+            deserializing_struct_key: false,
         }
     }
 
@@ -214,6 +218,16 @@ where
     /// Don't accept the old enum format used by `serde_cbor` versions <= v0.9.
     pub fn disable_legacy_enums(mut self) -> Self {
         self.accept_legacy_enums = false;
+        self
+    }
+
+    /// Set numeric index packed format starts with.
+    ///
+    /// By default, the packed format starts indexing fields
+    /// with 0. This allows deserializing packed formats that starting with 1,
+    /// or any other value small enough that field indices don't wrap.
+    pub fn packed_starts_with(mut self, offset: u8) -> Self {
+        self.packing_offset = offset;
         self
     }
 
@@ -258,7 +272,13 @@ where
 
     fn parse_u8(&mut self) -> Result<u8> {
         match self.next()? {
-            Some(byte) => Ok(byte),
+            Some(byte) => {
+                if self.deserializing_struct_key {
+                    Ok(byte - self.packing_offset)
+                } else {
+                    Ok(byte)
+                }
+            }
             None => Err(self.error(ErrorCode::EofWhileParsingValue)),
         }
     }
@@ -971,8 +991,11 @@ where
         *self.len -= 1;
 
         match self.de.peek()? {
-            Some(_byte @ 0x00..=0x1b) if !self.accept_packed => {
-                return Err(self.de.error(ErrorCode::WrongStructFormat));
+            Some(_byte @ 0x00..=0x1b) => {
+                if !self.accept_packed {
+                    return Err(self.de.error(ErrorCode::WrongStructFormat));
+                }
+                self.de.deserializing_struct_key = true;
             }
             Some(_byte @ 0x60..=0x7f) if !self.accept_named => {
                 return Err(self.de.error(ErrorCode::WrongStructFormat));
@@ -981,6 +1004,7 @@ where
         };
 
         let value = seed.deserialize(&mut *self.de)?;
+        self.de.deserializing_struct_key = false;
         Ok(Some(value))
     }
 
@@ -1022,8 +1046,11 @@ where
         K: de::DeserializeSeed<'de>,
     {
         match self.de.peek()? {
-            Some(_byte @ 0x00..=0x1b) if !self.accept_packed => {
-                return Err(self.de.error(ErrorCode::WrongStructFormat))
+            Some(_byte @ 0x00..=0x1b) => {
+                if !self.accept_packed {
+                    return Err(self.de.error(ErrorCode::WrongStructFormat));
+                }
+                self.de.deserializing_struct_key = true;
             }
             Some(_byte @ 0x60..=0x7f) if !self.accept_named => {
                 return Err(self.de.error(ErrorCode::WrongStructFormat))
@@ -1034,6 +1061,7 @@ where
         }
 
         let value = seed.deserialize(&mut *self.de)?;
+        self.de.deserializing_struct_key = false;
         Ok(Some(value))
     }
 
--- a/src/ser.rs
+++ b/src/ser.rs
@@ -51,6 +51,9 @@ where
 pub struct Serializer<W> {
     writer: W,
     packed: bool,
+    current_nesting_depth: u32,
+    pack_to_depth: u32,
+    packing_offset: u32,
     enum_as_map: bool,
 }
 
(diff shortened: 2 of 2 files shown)


nickray:
I force pushed to satisfy formatting checks.

Question: I'm completely stumped as to where the "packing index" gets checked in deserialization. Can you help?

pyfisch:
> I think packing a struct itself is not expressible via serde attributes, which is why it's implemented here?

"Packing a struct" (using integer values for the keys) is not expressible with serde. Therefore this crate keeps track of the field indizes and serializes them.
On the other hand serde happily accepts a struct in this format (and others) for deserialization.

>  Thoughts?

If there are just a few definitions it may be easiest to manually write the serializers and deserializers for them. This way you have full control over how the values are parsed and serialized. Another thing you need to watch out for is whether values are actually in canonical format. The only way to reliably check this with this crate is to serialize the parsed value again and compare it to the received CBOR document.

I think this is a really specific use-case and therefore it is unlikely I will include your changes in serde-cbor.

> Question: I'm completely stumped as to where the "packing index" gets checked in deserialization. Can you help?

See first answer.

nickray:
> "Packing a struct" (using integer values for the keys) is not expressible with serde.

Right, currently it isn't, but I see no reason why it couldn't. There's `rename` and `rename_all` attributes, mapping key strings to key strings. There could also be `map` or whatnot in general, mapping key strings to keys of a generic other type. Or specifically, something like `#[serde(index_from_zero)]` and `#[serde(index_from_one)]` attributes in particular that would replace the existing string keys in the serde data model maps with number keys, and back. I understand this would be a big discussion in serde.rs whether this is needed, or whether serialization formats should do this themselves - hence my patches to do this in the serialization format, CBOR.

It's *would* be much nicer to specify as much as possible in the type definition using attributes, than doing everything when performing serialization.

> If there are just a few definitions it may be easiest to manually write the serializers and deserializers for them. This way you have full control over how the values are parsed and serialized. Another thing you need to watch out for is whether values are actually in canonical format. The only way to reliably check this with this crate is to serialize the parsed value again and compare it to the received CBOR document.

I have quite a few (https://github.com/nickray/usbd-ctaphid/blob/main/src/types.rs)... This is an embedded use case, I don't check at run-time, but instead rely on acceptance tests to ensure all types are serialized as specified by the protocol (CTAP). In practice it's mostly a matter of defining the struct keys in the right order.

> I think this is a really specific use-case and therefore it is unlikely I will include your changes in serde-cbor.

I find it unfair to call the use of CBOR in WebAuthn/CTAP a "really specific use-case". After all, WebAuthn is a W3C standard. The indexing portion of the changes more generally have to do with whether one is communicating with another party that starts counting at 1 instead of 0 `¯\_(ツ)_/¯`. And the "depth" portion has to do with whether said other party is internally consistent in its thinking :'-). Both things are not in my hands.

--

I'd like to sum up the state of this (rather small) patch set:

- I have two requirements:
  a) ser/de with indices starting at 1 instead of 0
  b) ser/de using numerical indices only at top level, or top two levels, in nested structures
- Attached are patches that implement both serialization and deserialization, with configurable offset 0, 1, ... for a) and configurable depth for b). 
- This implementation is backwards-compatible for existing users of the library, so could even be released in a patch update.
- I would really not like to have to fork this library into some kind of `serde-ctap-cbor` when the requirements can easily be implemented here, the functionality is 99% the same, and i believe others will have these use cases too sooner or later (e.g./i.e. CTAP client = desktop use case, instead of my device = microcontroller case).

Can we shift the discussion to what would be necessary from the point of view of this repository's code standard to add the functionality, rather than discuss the need for it?

E.g. if say there are performance concerns about things like https://github.com/pyfisch/cbor/pull/168/files#diff-bcad83e135b6617746a5472d41ac63e9R276-R280, whether that should be feature-gated, etc.

pyfisch:
Hi @nickray,

I've thought a bit about your request and had a look at your code and the FIDO CTAP2 protcol.

First about the CTAP2 protocol: The FIDO protocol deals with authentication and therefore it is critical to get the implementation right and not introduce any security vulnerabilities. They have a defined a CBOR-based protocol with a required canonical encoding. Any implementation SHALL only emit the canonical encoding and SHOULD likewise only accept the canonical encoding. I believe if an implementation accepts more than the canonical encoding it exposes itself to a host of issues like duplicate keys and potential security issues from other kinds of malformed messages. One interesting feature of the protocol is that it uses both integers and strings as map keys. The top-level parameters use one-based integers while the other types like `PublicKeyCredentialRpEntity` are represented as maps with string keys.

About serde itself: Serde works with a wide variety of formats with quite different features. To enable this you give up some control about serialization and deserialization. For serialization you usually can't choose how exactly the document is formatted, just that it contains the necessary information. For deserialization serde accepts at least three formats for structures.

Take this struct for example:

```rust
struct Color {
    r: u8,
    g: u8,
    b: u8,
}
```

Both JSON documents `{"r": 120, "g": 42, "b": 0}` and `[120, 42, 0]` deserialize to the same struct even while one is a map and the other is an array. The third format is `{0: 120, 1: 42, 2: 0}` which is not expressible in JSON but is the packed format of CBOR. [**Playground Example**](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=1af2e2ca6ca9ef2b89b1a1009612491f)

Therefore I don't think serde is generally a good choice if you need to work with a strict canonical format of some sort. (For serde-cbor specifically it is possible to deserialize a document to `Value` and serialize the `Value` again. If the document and the serialized value match the document was in the canonical form from [RFC 7049 bis](https://tools.ietf.org/html/draft-ietf-cbor-7049bis-04#section-2). But this requires std and an extra serialization step.)

Because I haven't done any work on embedded Rust myself I will not comment to much on your CTAP2 code. You use the `heapless` crate to create fixed size buffers for strings and byte arrays. They are usually 64 elements long. This means with shorter values you waste the rest of each buffer while 64 bytes may be too short to store some parameters such as an URL. A better solution in my opinion would be to use `&[u8]` and `&str` instead. They would point to the stored message. (To construct messages you would need an arena allocator backed by a fixed buffer or similar to store the bytes and text.)

By the way I have recently written a no-std CBOR tokenizer that parses CBOR to a token stream. It gives very much control about how you deserialize a document but it requires writing a lot of code to deserialize to a struct. The code is at <https://github.com/pyfisch/minicbor/>. This should be quite useful to write a CBOR deserializer for an embedded device. (It right now lacks a way to check that an integer is in shortest possible form, but this can be easily added.) If you have questions how to use it, write an issue over there. On the other hand CBOR is *really* easy to parse so it should be easy enough to write a parser and serializer specifically for CTAP messages.

> I have two requirements:
> a) ser/de with indices starting at 1 instead of 0
> b) ser/de using numerical indices only at top level, or top two levels, in nested structures

I have strong reservations against using serde-cbor for the reasons mentioned above. Nevertheless it is possible to satisfy the two mentioned requirements without changes to serde-cbor or serde. Write a procedural macro like #[ctap(numbered_offset_one)] and attach it to every message you where you want indices starting at 1. This macro should generate the correct code to handle (de)serialization.

What do you think about this?

nickray:
I'll try the proc macro approach, will give me a good reason to learn them... Thanks for sharing your thoughts.

nickray:
Just to follow up on this, I did indeed write a proc-macro: https://github.com/nickray/serde-indexed

PR #172, opened 2020-01-06
Add support for CBOR tags

rklaehn:
Using a thread local value as a way to get the tag value over the fence from the CBOR format to a CBOR aware serializer/deserializer.

Replaces https://github.com/pyfisch/cbor/pull/164

Here is a more complex real world use case from @vmx : https://github.com/vmx/rust-ipld-dag-cbor


rklaehn:
As discussed in #164, the `tags` feature should not be enabled by default since it changes the deserialization behavior for non tag aware data structures when tags are present.

Not sure if `serde_cbor::Value::Tag` should exist when the feature is disabled.

pyfisch:
Thanks for all your work! :tada: 

I will release the new version to crates.io soon.

PR #174, opened 2020-01-10
Clarify packed format and add a to_vec_minimal helper method.

ctm:
This is the PR that includes the `to_vec_minimal` helper that you didn't ask for.

If you don't want it, I'll take it out and revert most of the packed encoding documentation, but add a paragraph or two that basically says the same thing as my Minimal Encoding paragraph, but instead of mentioning `to_vec_minimal` as part of serde-cbor I'll just show how the end-user can make such a beast.

I don't have a strong preference one way or the other. On one hand if `to_vec_minimal` is adopted and some other minimization options are added, then `to_vec_minimal` can be updated to use those techniques and everyone using `to_vec_minimal` automatically gets the added compression. On the other hand, every extra method in an API is some degree of burden on the maintainers.

### DIFF
--- a/src/ser.rs
+++ b/src/ser.rs
@@ -26,7 +26,7 @@ where
     Ok(vec)
 }
 
-/// Serializes a value to a vector in packed format.
+/// Serializes a value to a vector in packed format using externally tagged enums.
 #[cfg(feature = "std")]
 pub fn to_vec_packed<T>(value: &T) -> Result<Vec<u8>>
 where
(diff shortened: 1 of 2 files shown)


src/ser.rs
```diff
@@ -37,6 +37,17 @@ where
     Ok(vec)
 }
 
+/// Serializes a value to a vector in the most compressed format, albeit enums are not externally tagged.
+#[cfg(feature = "std")]
+pub fn to_vec_minimal<T>(value: &T) -> Result<Vec<u8>>
```
pyfisch: Unfortunately I don't want to add additional methods right now.

ctm:
No problem. I'll do a doc-only PR. However, if you'd like `legacy_enums` to go away, then I don't want to mention it, nor do I want to get Yew to use it.

pyfisch:
> if you'd like legacy_enums to go away

No I decided to keep it (mainly to read older files and write files to systems that expect the old format).

PR #175, opened 2020-01-10
Clarify packed format

ctm:
This PR tries to preserve as much of the spirit of the original Packed Encoding documentation as possible while still providing a warning that might be caught by people for whom the extra bytes of the enum variant may be a critical issue.

This PR does not show how to use `legacy_enums()` to get an even more compact representation, but I'm happy to add that info if desired. I don't want to add that info if `legacy_enums()` is likely to go away, especially if it will go away without there being some replacement that omits variant names from the encoding.


ctm:
I didn't see the update on the ticket. I see that `legacy_enums()` is not going away, so I will do a different PR.

PR #180, opened 2020-02-15
Use `serde::ser::StdError` shim type rather than `std::error::Error`

cbeck88:
This is for `alloc` vs. `std` modes compatibility, a code comment
explains in detail the purpose of this, which is admittedly subtle.

---

I'm evaluating using cbor in a large
project, I think a patch like this would be super helpful to us.

The gist of it is, we have a bunch of SGX enclaves written in rust,
and servers that pass messages to them in rust. I would like to use
`cbor/alloc` in the enclaves to deserialize messages across the FFI
boundary.

It would be great if `no_std` infra crates in our codebase
can unconditionally select `cbor/alloc` without the `std` flag (to minimize
configuration), and these crates can be consumed by servers that also
rely on third party crates from crates.io that may enable
`serde/std`. I think this patch will make that possible, I'm going to test it.


stevenroose:
It looks like `serde_json` is instead using `serde::de::StdError`: https://github.com/serde-rs/json/blob/4bcc3625c34b4f9b6c76013b8a5afcaed5a1f1da/src/error.rs#L317

It probably is not a meaningful difference.

stevenroose:
What is a meaningful difference though is that `serde_json` is only providing the `source` method on the `std` feature. Instead of only doing that for the `Error::Io` variant.

Issue #179, opened 2020-02-01
Looking for Maintainers


pyfisch:
Hi everyone,

when I started the `serde_cbor` crate in September 2015 it was a quick experiment on how to implement a new data format in serde. In the years since then the crate was much improved and it now appears to be used in quite a few projects. 

I'd like to thank all contributors including @chrysn (no-std and IO reader), @sfackler (a near complete rewrite in 2017 and several other changes), @baloo, @wildarch (no-std support), @rklaehn (tags) and [many more](https://github.com/pyfisch/cbor/graphs/contributors). In this time I have been the person who replied to many issues, did small fixes and reviewed most pull requests.

The crate has a bunch of build flags (cargo features) including `std`, `alloc` (like std but without io) and `tags`. CBOR documents can both be (de-)serialized using `Read`/`Write` and slices/`Vec<u8>`. At run-time the user can choose between legacy-enums or standard serde enum and packed encoding in addition to the [options offered by serde](https://serde.rs/attributes.html). Many issues discovered by users were due to a bad interaction between multiple features. Changes to the crate sometimes resulted in problems in a seemingly unrelated location. Within the past year it became increasingly clear to me that I don't understand all the details and interactions of different features. This is in part because I accepted pull requests implementing new features but did not review the code thoroughly enough. Some people expect me to add new features they need for their own project. If I reject these features because they complicate the crate and I don't want to maintain them they are unhappy.

I currently don't use CBOR in any project and therefore I maintained the crate mainly as a service to the Rust community. As serialization is not very exciting I kind of lost interest and often people need to wait a long time before I reply to their issues or pull requests.

For these reasons I decided not to continue the development of this crate and only occasionally review pull requests with bug fixes.

Unless of course a Rustacean who wants to continue the development and maintain the crate. There appears to be demand for a good CBOR implementation in Rust but currently this crate does not fit all use cases. To make the crate more useful and bug free I believe an almost complete rewrite is necessary.

The new project should consist of two parts. 
First a low-level crate that parses and writes CBOR but does not depend on serde. It should support reading all valid CBOR documents including those that contain tags and unknown simple values. Users should be able to choose all serialization details including the bit-width of integers and floats. This enables the implementation of custom serialization, validation and "canonical" encodings. 
[...]

PR #184, opened 2020-03-21
Create a workspace with serde_cbor as a member.

ahmedcharles:


stevenroose:
What is the motivation for this change? Without any other package in the top-level it doesn't make much sense to me to do this.

ahmedcharles:
The context is the discussion in #179. I have other changes locally which build on this, but they are perhaps more controversial, so I figured I'd make a fairly simple change which would help avoid merge conflicts and show incremental progress. But there's no interest...

stevenroose:
I don't see the idea of a workspace mentioned in #179 at all...

ahmedcharles:
It's implied by the mention of splitting the crate into a low level crate and one for just the serde integration.

hansl:
Please see my fork and branch. There is a workspace there, but there's no need to have this here as a stepping stone.

ahmedcharles:
I did look at your branch. It seems more like a rewrite than an incremental improvement. The primary reason I wanted to merge this is because I'd prefer reasoning about changes as incremental improvements rather than being forced to commit to one huge change all at once.

It seems like there isn't any interest in that, however.

PR #201, opened 2020-07-06
Add the cbor!() macro

npmccallum:
The cbor!() macro is inspired by serde_json::json!(), but the
implementation is simpler and the macro should be more useful. In
particular, the cbor!() macro returns Result<Value, Error>, so it can be
used in places that need careful error handling.


npmccallum:
One decision I was on the fence about: `cbor!(null)` or `cbor!(Null)`. The first one is closer to JSON and matches other keywords like `true` and `false`. The second one is closer to the Rust name. I could go either way. Let me know your preference.

pyfisch:
This could be a separate crate, right?

PR #207, opened 2020-08-03
Allow integer map keys to accept string-encoded integers

### LINKED ISSUE #205: Deserialisation of maps does not parse string keys to integers
This is one of those issues where it's not exactly clear what the correct thing to do is, but I feel like CBOR is mostly used as a compact version of JSON, so I think you should be able to write JSON using `serde_json`, convert it to CBOR, read it in using `serde_cbor` and it should give the same result.

This doesn't work for something like `HashMap<u32, String>`. The only key type that JSON supports is `String`, so when `serde_json` reads or writes that type it converts the integers to/from strings. The code to do this is [`MapKey`](https://github.com/serde-rs/json/blob/master/src/de.rs#L2104). `serde_cbor` doesn't have the same code, presumably because CBOR does support integer keys.

The CBOR authors did consider this issue. The spec has [this](https://tools.ietf.org/html/rfc7049#section-3.7) to say about map keys:

>    The encoding and decoding applications need to agree on what types of
   keys are going to be used in maps.  In applications that need to
   interwork with JSON-based applications, keys probably should be
   limited to UTF-8 strings only; otherwise, there has to be a specified
   mapping from the other CBOR types to Unicode characters, and this
   often leads to implementation errors.  In applications where keys are
   numeric in nature and numeric ordering of keys is important to the
   application, directly using the numbers for the keys is useful.
>
>   If multiple types of keys are to be used, consideration should be
   given to how these types would be represented in the specific
   programming environments that are to be used.  For example, in
   JavaScript objects, a key of integer 1 cannot be distinguished from a
   key of string "1".  This means that, if integer keys are used, the
   simultaneous use of string keys that look like numbers needs to be
   avoided.  Again, this leads to the conclusion that keys should be of
   a single CBOR type.

Timmmm:
This fixes #205. I know this project is semi-abandoned and you don't want to add code you haven't read (totally understandable) so let me know if you think this will never be merged and in that case I guess we will have to fork this.

Note most of the code for `impl<'de, 'a, R> de::Deserializer<'de> for MapKey<'a, R>` was copied from `serde_json`.

When serde attempts to deserialise a HashMap<i32, T> for example, with serde_json it will accept keys that are string-encoded versions of integers. This is because JSON only supports string keys. However this means that when you save data using serde_json, convert it to CBOR and then try and read it using serde_cbor, it will fail because CBOR does not accept string-encoded integer keys.

This commit changes that so that it does.

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -972,6 +972,160 @@ where
     }
 }
 
+// A wrapper around `visitor` to convert strings to integers. An version of this
+// struct is created for each integer type `$numtype` (e.g. u16). Any call to
+// `visit_str()` for this visitor will attempt to parse it to `$numtype` and pass
+// call `visit_$numtype()` on the wrapped visitor. Calls to `visit_$numtype()` are
+// passed through directly. All other visits result in an error.
+macro_rules! int_or_string_visitor {
+    ($visitor_name:ident, $visit:ident, $numtype:ty) => {
+        struct $visitor_name<'de, V: de::Visitor<'de>> {
+            inner_visitor: V,
+            _phantom_data: std::marker::PhantomData<&'de ()>,
+        }
+
+        impl<'de, V: de::Visitor<'de>> de::Visitor<'_> for $visitor_name<'de, V> {
+            type Value = V::Value;
+
+            fn expecting(&self, formatter: &mut std::fmt::Formatter) -> std::fmt::Result {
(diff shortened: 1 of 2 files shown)


Timmmm:
Also I'm not exactly sure what `accept_packed`/`accept_named` are so this might not be right if either of those are disabled.

PR #213, opened 2021-03-31
Implement `de::from_slice_count`

ericmarkmartin:
Add a function that deserializes first valid cbor value out of a slice of bytes
and returns the number of consumed bytes.

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -54,6 +54,37 @@ where
     Ok(value)
 }
 
+/// Decodes a value from the first piece of valid CBOR data in a slice.
+///
+/// # Examples
+///
+/// Deserialize a `String`
+///
+/// ```
+/// # use serde_cbor::de;
+/// let v: Vec<u8> = vec![0x66, 0x66, 0x6f, 0x6f, 0x62, 0x61, 0x72,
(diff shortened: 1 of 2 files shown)


ericmarkmartin:
@firstyear and I would really like this functionality for our work in kanidm/webauthn-rs, where we need to parse cbor values out of byte streams.

ericmarkmartin:
I'm also somewhat curious about why I have to subtract one off of the error's offset.

dtolnay (commented):
This use case doesn't strike me as warranting a dedicated API inside serde_cbor. I would prefer to recommend that the caller uses the following code instead, which I find simple/clear.

```rust
let mut deserializer = Deserializer::from_slice(slice);
let value = T::deserialize(&mut deserializer)?;
let processed = deserializer.byte_offset();
```

Firstyear:
Is there somewhere within the serde_cbor codebase where this could be added as an example usage?

ericmarkmartin:
> Is there somewhere within the serde_cbor codebase where this could be added as an example usage?

Maybe in `from_slice`? @dtolnay would you accept a PR adding such an example to the doc string?

